PR #30, opened 2023-04-08
Simplify OpenSubsonic API

Tolriq:
Since we only target to have a v1 API based on extension, we can simplify the API surface.


deluan (approved):

spl0k (approved):

epoupon (approved):

deluan (approved):

Tolriq:
So no rebase after the reviews even from Github UI :) Good to know.

PR #36, opened 2023-04-13
Add Code of Conduct

deluan:
We still need to put in place a way for reporting abuse, see below.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,133 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the overall
+  community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or advances of
+  any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,133 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+[INSERT CONTACT METHOD].
```
deluan, line 64: This should be an email or any other of anonymous reporting communication channel. Any ideas?

Tolriq: Since we are for the moment 4 "owners" and it could concern one of us too, does it really need to be anonymous? An email is not really anonymous unless the user create a fake account just for the report anyway and he could create a throwaway account on Github too.
That would require one more service to maintain.
Or maybe there's integrated tools in Github?

deluan: Sorry, I meant _private_, not _anonymous_. 

> Or maybe there's integrated tools in Github?

That's what I'm asking, if anyone knows any tool that could be used for this.

Tolriq: Well we can use the private security report then. After all harassment is a security concern no?

deluan: Not sure... I think it may be better to simply list the email of all owners. Whoever receives it has the responsibility to bring the issue to the group.

deluan: Relevant to this: https://github.com/badges/shields/issues/2671#issuecomment-459901586

Tolriq: Then we'd need a place to discuss it as a 4 people email chain is horrible :( let's find a tool then. Alternative would be a free slack instance so people can pm us there and we can have a channel to discuss the report ?

deluan: Found it. GitHub has a tool for that: https://docs.github.com/en/communities/moderating-comments-and-conversations/managing-how-contributors-report-abuse-in-your-organizations-repository

Screenshot 2023-04-13 at 4 30 48 PM

Should I enable it, and change the doc to reflect it?

deluan: I'd like to have the opinion from other members as well. 
cc: @opensubsonic/clients @opensubsonic/servers

epoupon: > Found it. GitHub has a tool for that: https://docs.github.com/en/communities/moderating-comments-and-conversations/managing-how-contributors-report-abuse-in-your-organizations-repository
> 
> Screenshot 2023-04-13 at 4 30 48 PM
> 
> Should I enable it, and change the doc to reflect it?

I would say yes. We can open it to every github users and restrict later if people abuse the reporting system

austinried: LGTM

dweymouth (approved):

PR #38, opened 2023-04-15
Clarify id parameter of getCoverArt

Tolriq:
The id parameter is not the id of the media, it's the covertArt ID, a value that is returned by many endpoints via the covertArt property.

### DIFF
--- a/content/en/docs/Endpoints/getcoverart.md
+++ b/content/en/docs/Endpoints/getcoverart.md
@@ -15,7 +15,7 @@ Returns a cover art image.
 
 | Parameter | Req. | OpenS. | Default | Comment |
 | --- | --- | --- | --- | --- |
-| `id` | **Yes** |     | The ID of a song, album or artist. |
+| `id` | **Yes** |     | The coverArt ID. Returned by most entities likes [`Child`](../../responses/child) or [`AlbumID3`](../../responses/albumid3) |
 | `size` | No  |     | If specified, scale image to this size. |
 
 ### Example


content/en/docs/Endpoints/getcoverart.md
```diff
@@ -15,7 +15,7 @@ Returns a cover art image.
 
 | Parameter | Req. | OpenS. | Default | Comment |
 | --- | --- | --- | --- | --- |
-| `id` | **Yes** |     | The ID of a song, album or artist. |
+| `id` | **Yes** |     | The covertArt ID. Returned by most entities likes [`Child`](../../responses/child) or [`AlbumID3`](../../responses/albumid3) |
```
epoupon, line 18: typo: coverArt

epoupon (approved):

deluan (commented):
This is a breaking change. A few clients out there use the ID of the resource (album, song, artist, playlist) to retrieve covers even if the `coverArt` property is empty. Pretty sure [Airsonix-Refix](https://github.com/tamland/airsonic-refix) and DSub do this.

Tolriq:
Not it's not they use well known covertArt IDs like ar-id or al-id or pl-id those are covertArt IDs that happens to have the item id but also have a prefix so they are not the actual IDs as the docs suggest.
So this does not change anything. They can keep using well known ids if they want to abuse the system and servers can keep supporting them if they want.
The clarification is to makes things more clear and avoid the request about universal unique IDs that are not needed as even the well known ids handle this.

austinried:
I think it still is a breaking change though, even if those clients in particular are using a well-known cover art ID instead of the coverArt value. Due to the way it's originally worded, many servers also have support for the literal entity ID, and re-wording it to get rid of that support would be breaking for any clients depending on it.

Tolriq:
See my answer at https://github.com/opensubsonic/open-subsonic-api/discussions/35#discussioncomment-5627098 and let's move the discussion here. Validating a broken doc is also a breaking change, but with a lot more impacts.

lachlan-00:
Are these prefixes listed anywhere?

Tolriq:
No they are not officially documented. There's no official rules, hence the need to avoid confusion and more reaction from people in the discussing thread. If we can't get people to agree on the base stuff, it will be hard to move with the rest of the things.

lachlan-00:
if we defined a coverart prefix this seems like a no-brainer to me.

I have `al-` `ar-` `pl-` and `pod-` in my list of objects for getcoverart

Tolriq:
Well known prefix does not fix all of the api miss usage currently made.
Then comes what are the answer of the servers when there's no such art. Some servers returns a default cover.
IMO this is bad because this leave the choice out of the client to decide what to display when there's no art. 
Default covers from each servers may look perfectly ugly in some client UIs and themes.

The goal of OpenSubsonic is to clarify the API and clean up some mess when possible. We should not encourage bad behavior and properly handle / describe how details works to avoid issues.

Lack of coverArt returned for albums or artists for example would be a reason to call the getXXXInfo to get other possible arts.
With some servers always returning a covertArt even when there's none this prevent clients to do that.
With some servers always returning an image even when the covertArt does not exist this also prevent clients to do that.

To achieve the goal of the project we must clear friction points and things that leaves servers doing different things for basic features.
When in this case half the servers do A and the other half do B this will impact half of the things, we should aim to improve the situation and so choose to eventually break the bad case to encourage the good case. And not break the good case and force the good servers to behave badly for no reason.

deluan:
> Not it's not they use well known covertArt IDs like ar-id or al-id or pl-id those are covertArt IDs that happens to have the item id but also have a prefix so they are not the actual IDs as the docs suggest. So this does not change anything. They can keep using well known ids if they want to abuse the system and servers can keep supporting them if they want. The clarification is to makes things more clear and avoid the request about universal unique IDs that are not needed as even the well known ids handle this.

You are right! Sorry for the noise.

deluan (approved):

lachlan-00 (approved):

PR #51, opened 2023-09-27
Expose many new fields to clients

Tolriq:
Support for multi values fields like genre and artists. Support song contributors like Composers, ...
Support many new useful data for Songs, Artists and Albums.

### DIFF
--- a/content/en/docs/Responses/AlbumID3.md
+++ b/content/en/docs/Responses/AlbumID3.md
@@ -11,19 +11,66 @@ description: >
 {{< tab header="**Example**:" disabled=true />}}
 {{< tab header="OpenSubsonic" lang="json">}}
 {
-  "id": "ad0f112b6dcf83de5e9cae85d07f0d35",
-  "name": "8-bit lagerfeuer",
-  "artist": "pornophonique",
-  "year": 2007,
-  "coverArt": "al-ad0f112b6dcf83de5e9cae85d07f0d35_640a93a8",
-  "starred": "2023-03-22T01:51:06Z",
-  "duration": 1954,
-  "playCount": 97,
-  "created": "2023-03-10T02:19:35.784818075Z",
-  "artistId": "91c3901ac465b9efc439e4be4270c2b6",
-  "songCount": 8,
-  "played": "2023-03-28T00:45:13Z",
-  "userRating": 4
+    "id": "ad0f112b6dcf83de5e9cae85d07f0d35",
+    "name": "8-bit lagerfeuer",
+    "artist": "pornophonique",
+    "year": 2007,
+    "coverArt": "al-ad0f112b6dcf83de5e9cae85d07f0d35_640a93a8",
+    "starred": "2023-03-22T01:51:06Z",
+    "duration": 1954,
+    "playCount": 97,
+    "genre": "Hip-Hop",
+    "created": "2023-03-10T02:19:35.784818075Z",
+    "artistId": "91c3901ac465b9efc439e4be4270c2b6",
+    "songCount": 8,
+    "played": "2023-03-28T00:45:13Z",
+    "userRating": 4,
+    "recordLabels": [
+        {
+            "name": "Sony"
+        }
+    ],
+    "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
+    "genres": [
+        {
+            "name": "Hip-Hop"
+        },
+        {
+            "name": "East coast"
+        }
+    ],
+    "artists": [
+        {
+            "id": "ar-1",
+            "name": "Artist 1"
+        },
+        {
+            "id": "ar-2",
+            "name": "Artist 2"
+        }
+    ],
+    "displayArtist": "Artist 1 feat. Artist 2",
+    "releaseTypes": [
+        "Album",
+        "Remixes"
+    ],
+    "moods": [
+        "slow",
+        "cool"
+    ],
+    "sortName": "lagerfeuer (8-bit)",
+    "originalReleaseDate": "2001-03-10T02:19:35.784818075Z",
+    "isCompilation": false,
+    "discTitles": [
+        {
+            "disc": 0,
+            "title": "Disc 0 title"
+        },
+        {
+            "disc": 2,
+            "title": "Disc 1 title"
+        }
+    ]
 }
 {{< /tab >}}
 {{< tab header="Subsonic" lang="json" >}}
@@ -36,6 +83,7 @@ description: >
   "starred": "2023-03-22T01:51:06Z",
   "duration": 1954,
   "playCount": 97,
+  "genre": "Hip-Hop",
   "created": "2023-03-10T02:19:35.784818075Z",
   "artistId": "91c3901ac465b9efc439e4be4270c2b6",
   "songCount": 8
@@ -59,12 +107,38 @@ description: >
 | `genre` | `string` | No |     | The album genre|
 | `played` | `string` | No | **Yes**    | Date the album was last played. [ISO 8601]|
 | `userRating` | `int` | No | **Yes**    | The user rating of the album. [1-5]|
+| `recordLabels` | Array of [`RecordLabel`](../recordlabel) | No |  **Yes**   | The labels producing the album. |
+| `musicBrainzId` | `string` | No |  **Yes**   | The album MusicBrainzID. |
+| `genres` | Array of [`ItemGenre`](../itemgenre) | No | **Yes**    | The list of all genres of the album. |
+| `artists` | Array of [`ArtistID3`](../artistid3) | No | **Yes**    | The list of all album artists of the album. (Note: Only the required [`ArtistID3`](../artistid3) fields should be returned by default)|
+| `displayArtist` | `string` | No |  **Yes**   | The single value display artist. |
+| `releaseTypes` | Array of `string` | No | **Yes**    | The types of this album release. (Album, Compilation, EP, Remix, ...).|
+| `moods` | Array of `string` | No | **Yes**    | The list of all moods of the album. |
+| `sortName` | `string` | No |  **Yes**   | The album sort name. |
+| `originalReleaseDate` | `string` | No |   **Yes**   | Date the album was originally released. [ISO 8601]|
+| `isCompilation` | `boolean` | No |  **Yes**    | True if the album is a compilation. |
+| `discTitles` | Array of [`DiscTitle`](../disctitle) | No | **Yes**    | The list of all disc titles of the album. |
 
 {{< alert color="warning" title="OpenSubsonic" >}}
 New fields are added:
 
 - `played`
 - `userRating`
+- `recordLabels`
+- `musicBrainzId`
+- `genres`
+- `artists`
+- `displayArtist`
+- `releaseType`
+- `moods`
+- `sortName`
+- `originalReleaseDate`
+- `isCompilation`
+- `discTitles`
+
+**Note**: All OpenSubsonic added fields are **optionals**. But if a server support a field it **must** return it with an empty / default value when not present in it's database so that clients knows what the server supports.
+
+**Note**: Even if some added fields may looks duplicated, it's important to still return the legacy data for compatibility reasons.
 {{< /alert >}}
 
 ---
--- a/content/en/docs/Responses/AlbumID3WithSongs.md
+++ b/content/en/docs/Responses/AlbumID3WithSongs.md
@@ -11,69 +11,114 @@ description: >
 {{< tab header="**Example**:" disabled=true />}}
 {{< tab header="OpenSubsonic" lang="json">}}
 {
-  "id": "200000021",
-  "parent": "100000036",
-  "album": "Forget and Remember",
-  "title": "Forget and Remember",
-  "name": "Forget and Remember",
-  "isDir": true,
-  "coverArt": "al-200000021",
-  "songCount": 20,
-  "created": "2021-07-22T02:09:31+00:00",
-  "duration": 4248,
-  "playCount": 0,
-  "artistId": "100000036",
-  "artist": "Comfort Fit",
-  "year": 2005,
-  "genre": "Hip-Hop",
-  "song": [
-    {
-      "id": "300000116",
-      "parent": "200000021",
-      "title": "Can I Help U?",
-      "isDir": false,
-      "isVideo": false,
-      "type": "music",
-      "albumId": "200000021",
-      "album": "Forget and Remember",
-      "artistId": "100000036",
-      "artist": "Comfort Fit",
-      "coverArt": "300000116",
-      "duration": 103,
-      "bitRate": 216,
-      "track": 1,
-      "year": 2005,
-      "genre": "Hip-Hop",
-      "size": 2811819,
-      "discNumber": 1,
-      "suffix": "mp3",
-      "contentType": "audio/mpeg",
-      "path": "user/Comfort Fit/Forget And Remember/1 - Can I Help U?.mp3"
-    },
-    {
-      "id": "300000121",
-      "parent": "200000021",
-      "title": "Planetary Picknick",
-      "isDir": false,
-      "isVideo": false,
-      "type": "music",
-      "albumId": "200000021",
-      "album": "Forget and Remember",
-      "artistId": "100000036",
-      "artist": "Comfort Fit",
-      "coverArt": "300000121",
-      "duration": 358,
-      "bitRate": 238,
-      "track": 2,
-      "year": 2005,
-      "genre": "Hip-Hop",
-      "size": 10715592,
-      "discNumber": 1,
-      "suffix": "mp3",
-      "contentType": "audio/mpeg",
-      "path": "user/Comfort Fit/Forget And Remember/2 - Planetary Picknick.mp3"
-    }
-  ]
+    "id": "ad0f112b6dcf83de5e9cae85d07f0d35",
+    "name": "8-bit lagerfeuer",
+    "artist": "pornophonique",
+    "year": 2007,
+    "coverArt": "al-ad0f112b6dcf83de5e9cae85d07f0d35_640a93a8",
+    "starred": "2023-03-22T01:51:06Z",
+    "duration": 1954,
+    "playCount": 97,
+    "genre": "Hip-Hop",
+    "created": "2023-03-10T02:19:35.784818075Z",
+    "artistId": "91c3901ac465b9efc439e4be4270c2b6",
+    "songCount": 8,
+    "played": "2023-03-28T00:45:13Z",
+    "userRating": 4,
+    "recordLabels": [
+        {
+            "name": "Sony"
+        }
+    ],
+    "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
+    "genres": [
+        {
+            "name": "Hip-Hop"
+        },
+        {
+            "name": "East coast"
+        }
+    ],
+    "artists": [
+        {
+            "id": "ar-1",
+            "name": "Artist 1"
+        },
+        {
+            "id": "ar-2",
+            "name": "Artist 2"
+        }
+    ],
+    "displayArtist": "Artist 1 feat. Artist 2",
+    "releaseTypes": [
+        "Album",
+        "Remixes"
+    ],
+    "moods": [
+        "slow",
+        "cool"
+    ],
+    "sortName": "lagerfeuer (8-bit)",
+    "originalReleaseDate": "2001-03-10T02:19:35.784818075Z",
+    "isCompilation": false,
+    "discTitles": [
+        {
+            "disc": 0,
+            "title": "Disc 0 title"
+        },
+        {
+            "disc": 2,
+            "title": "Disc 1 title"
+        }
+    ],
+    "song": [
+        {
+            "id": "300000116",
+            "parent": "200000021",
+            "title": "Can I Help U?",
+            "isDir": false,
+            "isVideo": false,
+            "type": "music",
+            "albumId": "200000021",
+            "album": "Forget and Remember",
+            "artistId": "100000036",
+            "artist": "Comfort Fit",
+            "coverArt": "300000116",
+            "duration": 103,
+            "bitRate": 216,
+            "track": 1,
+            "year": 2005,
+            "genre": "Hip-Hop",
+            "size": 2811819,
+            "discNumber": 1,
+            "suffix": "mp3",
+            "contentType": "audio/mpeg",
+            "path": "user/Comfort Fit/Forget And Remember/1 - Can I Help U?.mp3"
+        },
+        {
+            "id": "300000121",
+            "parent": "200000021",
+            "title": "Planetary Picknick",
+            "isDir": false,
+            "isVideo": false,
+            "type": "music",
+            "albumId": "200000021",
+            "album": "Forget and Remember",
+            "artistId": "100000036",
+            "artist": "Comfort Fit",
+            "coverArt": "300000121",
+            "duration": 358,
+            "bitRate": 238,
+            "track": 2,
+            "year": 2005,
+            "genre": "Hip-Hop",
+            "size": 10715592,
+            "discNumber": 1,
+            "suffix": "mp3",
+            "contentType": "audio/mpeg",
+            "path": "user/Comfort Fit/Forget And Remember/2 - Planetary Picknick.mp3"
+        }
+    ]
 }
 {{< /tab >}}
 {{< tab header="Subsonic" lang="json" >}}
@@ -161,6 +206,17 @@ description: >
 | `genre` | `string` | No |     | The album genre|
 | `played` | `string` | No | **Yes**    | Date the album was last played. [ISO 8601]|
 | `userRating` | `int` | No | **Yes**    | The user rating of the album. [1-5]|
+| `recordLabels` | Array of [`RecordLabel`](../recordlabel) | No |  **Yes**   | The labels producing the album. |
+| `musicBrainzId` | `string` | No |  **Yes**   | The album MusicBrainzID. |
+| `genres` | Array of [`ItemGenre`](../itemgenre) | No | **Yes**    | The list of all genres of the album. |
+| `artists` | Array of [`ArtistID3`](../artistid3) | No | **Yes**    | The list of all album artists of the album. |
+| `displayArtist` | `string` | No |  **Yes**   | The single value display artist. |
+| `releaseTypes` | Array of `string` | No | **Yes**    | The types of this album release. (Album, Compilation, EP, Remix, ...).|
+| `moods` | Array of `string` | No | **Yes**    | The list of all moods of the album. |
+| `sortName` | `string` | No |  **Yes**   | The album sort name. |
+| `originalReleaseDate` | `string` | No |   **Yes**   | Date the album was originally released. [ISO 8601]|
+| `isCompilation` | `boolean` | No |  **Yes**    | True if the album is a compilation.|
+| `discTitles` | Array of [`DiscTitle`](../disctitle) | No | **Yes**    | The list of all disc titles of the album. |
 | `song` | Array of [`Child`](../child) | No |     | The list of songs |
 
 {{< alert color="warning" title="OpenSubsonic" >}}
@@ -168,4 +224,26 @@ New fields are added:
 
 - `played`
 - `userRating`
+- `recordLabels`
+- `musicBrainzId`
+- `genres`
+- `artists`
+- `displayArtist`
+- `releaseType`
+- `moods`
+- `sortName`
+- `originalReleaseDate`
+- `isCompilation`
+- `discTitles`
+
+**Note**: All OpenSubsonic added fields are **optionals**. But if a server support a field it **must** return it with an empty / default value when not present in it's database so that clients knows what the server supports.
+
+**Note**: Even if some added fields may looks duplicated, it's important to still return the legacy data for compatibility reasons.
+{{< /alert >}}
+
+{{< alert color="warning" title="OpenSubsonic server support" >}}
+
+| Server | Min vers. | Comment |
+| --- | --- | --- |
+
 {{< /alert >}}
--- a/content/en/docs/Responses/ArtistID3.md
+++ b/content/en/docs/Responses/ArtistID3.md
@@ -1,6 +1,8 @@
 ---
 title: "ArtistID3"
-linkTitle: "ArtistID3"
+linkTitle: "ArtistID3 [OS]"
+opensubsonic:
+- Change
 description: >
   An artist from ID3 tags.
 ---
@@ -9,13 +11,20 @@ description: >
 {{< tab header="**Example**:" disabled=true />}}
 {{< tab header="OpenSubsonic" lang="json">}}
 {
-  "id": "37ec820ca7193e17040c98f7da7c4b51",
-  "name": "2 Mello",
-  "coverArt": "ar-37ec820ca7193e17040c98f7da7c4b51_0",
-  "albumCount": 1,
-  "userRating": 5,
-  "artistImageUrl": "https://demo.org/image.jpg",
-  "starred": "2017-04-11T10:42:50.842Z"
+    "id": "37ec820ca7193e17040c98f7da7c4b51",
+    "name": "2 Mello",
+    "coverArt": "ar-37ec820ca7193e17040c98f7da7c4b51_0",
+    "albumCount": 1,
+    "userRating": 5,
+    "artistImageUrl": "https://demo.org/image.jpg",
+    "starred": "2017-04-11T10:42:50.842Z",
+    "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
+    "sortName": "Mello (2)",
+    "roles": [
+        "artist",
+        "albumartist",
+        "composer"
+    ]
 }
 {{< /tab >}}
 {{< tab header="Subsonic" lang="json" >}}
@@ -39,3 +48,26 @@ description: >
 | `artistImageUrl` | `string` | No  |   | An url to an external image source. |
 | `albumCount` | `int` | No |     | Artist album count.  |
 | `starred` | `string` | No |     | Date the artist was starred. [ISO 8601]|
+| `musicBrainzId` | `string` | No |  **Yes**   | The artist MusicBrainzID. |
+| `sortName` | `string` | No |  **Yes**   | The artist sort name. |
+| `roles` | Array of `string` | No | **Yes**    | The list of all roles this artist have in the library. |
+
+{{< alert color="warning" title="OpenSubsonic" >}}
+New fields are added:
+
+- `musicBrainzId`
+- `sortName`
+- `roles`
+
+**Note**: All OpenSubsonic added fields are **optionals**. But if a server support a field it **must** return it with an empty / default value when not present in it's database so that clients knows what the server supports.
+
+**Note**: Even if some added fields may looks duplicated, it's important to still return the legacy data for compatibility reasons.
+{{< /alert >}}
+
+---
+
+{{< alert color="warning" title="OpenSubsonic server support" >}}
+
+| Server | Min vers. | Comment |
+| --- | --- | --- |
+{{< /alert >}}
--- a/content/en/docs/Responses/Child.md
+++ b/content/en/docs/Responses/Child.md
@@ -11,30 +11,108 @@ description: >
 {{< tab header="**Example**:" disabled=true />}}
 {{< tab header="OpenSubsonic" lang="json">}}
 {
-  "id": "082f435a363c32c57d5edb6a678a28d4",
-  "parent": "e8a0685e3f3ec6f251649af2b58b8617",
-  "isDir": false,
-  "title": "\"polar expedition\"",
-  "album": "Live at The Casbah - 2005-04-29",
-  "artist": "The New Deal",
-  "track": 4,
-  "year": 2005,
-  "coverArt": "mf-082f435a363c32c57d5edb6a678a28d4_6410b3ce",
-  "size": 19866778,
-  "contentType": "audio/flac",
-  "suffix": "flac",
-  "starred": "2023-03-27T09:45:27Z",
-  "duration": 178,
-  "bitRate": 880,
-  "path": "The New Deal/Live at The Casbah - 2005-04-29/04 - \"polar expedition\".flac",
-  "playCount": 8,
-  "played": "2023-03-26T22:27:46Z",
-  "discNumber": 1,
-  "created": "2023-03-14T17:51:22.112827504Z",
-  "albumId": "e8a0685e3f3ec6f251649af2b58b8617",
-  "artistId": "97e0398acf63f9fb930d7d4ce209a52b",
-  "type": "music",
-  "isVideo": false
+    "id": "082f435a363c32c57d5edb6a678a28d4",
+    "parent": "e8a0685e3f3ec6f251649af2b58b8617",
+    "isDir": false,
+    "title": "\"polar expedition\"",
+    "album": "Live at The Casbah - 2005-04-29",
+    "artist": "The New Deal",
+    "track": 4,
+    "year": 2005,
+    "coverArt": "mf-082f435a363c32c57d5edb6a678a28d4_6410b3ce",
+    "size": 19866778,
+    "contentType": "audio/flac",
+    "suffix": "flac",
+    "starred": "2023-03-27T09:45:27Z",
+    "duration": 178,
+    "bitRate": 880,
+    "path": "The New Deal/Live at The Casbah - 2005-04-29/04 - \"polar expedition\".flac",
+    "playCount": 8,
+    "played": "2023-03-26T22:27:46Z",
+    "discNumber": 1,
+    "created": "2023-03-14T17:51:22.112827504Z",
+    "albumId": "e8a0685e3f3ec6f251649af2b58b8617",
+    "artistId": "97e0398acf63f9fb930d7d4ce209a52b",
+    "type": "music",
+    "mediaType": "song",
+    "isVideo": false,
+    "bpm": 134,
+    "comment": "This is a song comment",
+    "sortName": "Polar expedition",
+    "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
+    "genres": [
+        {
+            "name": "Hip-Hop"
+        },
+        {
+            "name": "East coast"
+        }
+    ],
+    "artists": [
+        {
+            "id": "ar-1",
+            "name": "Artist 1"
+        },
+        {
+            "id": "ar-2",
+            "name": "Artist 2"
+        }
+    ],
+    "displayArtist": "Artist 1 feat. Artist 2",
+    "albumArtists": [
+        {
+            "id": "ar-6",
+            "name": "Artist 6"
+        },
+        {
+            "id": "ar-7",
+            "name": "Artist 7"
+        }
+    ],
+    "displayAlbumArtist": "Artist 6 & Artist 7",
+    "contributors": [
+        {
+            "role": "composer",
+            "artist": {
+                "id": "ar-3",
+                "name": "Artist 3"
+            }
+        },
+        {
+            "role": "composer",
+            "artist": {
+                "id": "ar-4",
+                "name": "Artist 4"
+            }
+        },
+        {
+            "role": "lyricist",
+            "artist": {
+                "id": "ar-5",
+                "name": "Artist 5"
+            }
+        },
+        {
+            "role": "performer",
+            "subRole": "Bass",
+            "artist": {
+                "id": "ar-5",
+                "name": "Artist 5"
+            }
+        }
+    ],
+    "displayComposer": "Artist 3, Artist 4",
+    "moods": [
+        "slow",
+        "cool"
+    ],
+    "replayGain": {
+        "trackGain": 0.1,
+        "albumGain": 1.1,
+        "trackPeak": 9.2,
+        "albumPeak": 9,
+        "baseGain": 0
+    }
 }
 {{< /tab >}}
 {{< tab header="Subsonic" lang="json" >}}
@@ -71,7 +149,7 @@ description: >
 | `id` | `string` | **Yes** |     | The id of the media |
 | `parent` | `string` | No |     | The id of the parent (folder/album) |
 | `isDir` | `boolean` | **Yes** |     | The media is a directory |
-| `title` | `string` | **Yes** |     | The medua name. |
+| `title` | `string` | **Yes** |     | The media name. |
 | `album` | `string` | No |     | The album name. |
 | `artist` | `string` | No |     | The artist name.  |
 | `track` | `int` | No |    | The track number. |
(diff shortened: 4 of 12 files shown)


content/en/docs/Responses/ReplayGain.md
```diff
@@ -0,0 +1,45 @@
...
+Does not exist.
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `trackGain` | `number` | No | **Yes**    | The track replay gain value. (In Db) |
+| `albumGain` | `number` | No | **Yes**    | The album replay gain value. (In Db) |
+| `trackPeak` | `number` | No | **Yes**    | The track peak value. (Range [0.0-1.0]) |
+| `albumPeak` | `number` | No | **Yes**    | The album peak value. (Range [0.0-1.0]) |
```
kgarner7, line 31: Small comment here, but the example fields `trackPeak`, `albumPeak` do not match the "Details" section, which limits it to a range of [0.0, 1.0]. The peak can [definitely exceed 1.0 though](https://wiki.hydrogenaud.io/index.php?title=ReplayGain_specification#Peak_amplitude), so I'd recommend updating the details section

Tolriq: @kgarner7 I must admit I only use the Gain values myself they are based on the comment from https://github.com/opensubsonic/open-subsonic-api/discussions/10 @dweymouth but yes from that doc it seems it can go higher even if not seen from Plex data for example.

content/en/docs/Responses/AlbumID3.md
```diff
@@ -11,19 +11,55 @@ description: >
...
+        {
+            "id": "ar-1",
+            "name": "Artist 1"
+        },
+        {
+            "id": "ar-2",
+            "name": "Artist 2"
+        }
+    ],
+    "displayArtist": "Artist 1 feat. Artist 2",
```
dweymouth, line 44: Do we need a `displayArtist` field? Legacy subsonic clients won't use any of the new fields and a modern OpenSubsonic client has the full artist info through `artists` and could come up with whatever display artist name they want (like Artist 1, Artist 2 with each being a clickable hyperlink)

content/en/docs/Responses/AlbumID3.md
```diff
@@ -11,19 +11,55 @@ description: >
...
+    "starred": "2023-03-22T01:51:06Z",
+    "duration": 1954,
+    "playCount": 97,
+    "genre": "Hip-Hop",
+    "created": "2023-03-10T02:19:35.784818075Z",
+    "artistId": "91c3901ac465b9efc439e4be4270c2b6",
+    "songCount": 8,
+    "played": "2023-03-28T00:45:13Z",
+    "userRating": 4,
+    "albumLabel": "Sony",
```
dweymouth, line 28: Maybe rename this to `recordLabel` or just `label`?

content/en/docs/Responses/ReplayGain.md
kgarner7: Makes sense! I do have some files of my own that are tagged with peak > 1.0, so I wanted to chime in here

content/en/docs/Responses/ReplayGain.md
```diff
@@ -0,0 +1,45 @@
+---
```
dweymouth, line 1: For replay gain it may be worth making a note somewhere that if a track is missing replayGain info, the server must return `"replayGain": {}` as the default and **not** `replayGain: { "trackGain": 0, "albumGain": 0, ... }`

Tolriq: This does not change much on client side handling, but since the fields are optional they by default should be omitted since it's a new field.

content/en/docs/Responses/AlbumID3.md
Tolriq: label might be missunderstood as a label, but  yes I can rename to recordLabel.

content/en/docs/Responses/Child.md
```diff
@@ -11,30 +11,95 @@ description: >
...
+        {
+            "id": "ar-1",
+            "name": "Artist 1"
+        },
+        {
+            "id": "ar-2",
+            "name": "Artist 2"
+        }
+    ],
+    "displayArtist": "Artist 1 feat. Artist 2",
```
sentriz, line 56: in the case of a Child as we are here, how does displayArtist different from artist? "The New Deal" in this case

content/en/docs/Responses/AlbumID3WithSongs.md
```diff
@@ -161,11 +195,44 @@ description: >
...
+- `albumLabel`
+- `musicBrainzId`
+- `genres`
+- `artists`
+- `displayArtist`
+- `releaseType`
+- `moods`
+- `sortName`
+- `originalReleaseDate`
+- `isCompilation`
```
sentriz, line 225: how do we specifically define complication here? can it be deduced from `len(artists) > 1`, or is it based on file metadata, where user may have an album which has multiple artists but still not a compilation?

content/en/docs/Responses/ReplayGain.md
dweymouth: It does change things on the client side. A trackGain or albumGain of 0 tells the client that no replayGain processing should be applied. Missing replayGain info means the client can use a default gain (like -8 dB, since most popular music has a replayGain value of around -8 to -10).

content/en/docs/Responses/ReplayGain.md
dweymouth: Yeah @kgarner7 is right - replayGain peaks of a bit over 1 are definitely possible. It was an oversight on my part in #10 to specify a [0..1] range.

content/en/docs/Responses/AlbumID3WithSongs.md
dweymouth: I think it's based on file metadata. There is a COMPILATION ID3 tag that Navidrome and iTunes and others use, to identify an album, say like "Top 90s Hits" that has a different artist for each track. (And the album artist is often tagged as Various Artists in this case)

content/en/docs/Responses/AlbumID3.md
```diff
@@ -59,12 +96,38 @@ description: >
 | `genre` | `string` | No |     | The album genre|
 | `played` | `string` | No | **Yes**    | Date the album was last played. [ISO 8601]|
 | `userRating` | `int` | No | **Yes**    | The user rating of the album. [1-5]|
+| `albumLabel` | `string` | No |  **Yes**   | The label producing the album. |
+| `musicBrainzId` | `string` | No |  **Yes**   | The album MusicBrainzID. |
+| `genres` | Array of `string` | No | **Yes**    | The list of all genres of the album. |
+| `artists` | Array of [`ArtistID3`](../artistid3) | No | **Yes**    | The list of all album artists of the album. (Note: Only the required [`ArtistID3`](../artistid3) fields should be returned by default)|
+| `displayArtist` | `string` | No |  **Yes**   | The single value display artist. |
+| `releaseType` | `string` | No | **Yes**    | The type of release. If the server handle subtypes then the field should return them type and subtypes separated with " / ".|
```
epoupon, line 104: Why not an array?
Should we try to normalize some possible values or at least give some examples?
Furthermore, I would expect to put here "Compilation" (like in https://musicbrainz.org/doc/Release_Group/Type)

content/en/docs/Responses/AlbumID3WithSongs.md
Tolriq: Yes there's a couple of tags for that, but I do not want to be more explicit as a server might offer a way for an user to set an album as a compilation manually.

content/en/docs/Responses/AlbumID3.md
Tolriq: It was discussed already, but to resume no other server do support multi fields, and the tag is single valued and can contain anything.
There's no point in forcing an array here for very little value gain. (Including impacts on the servers to split the value + normalize and everything, when every other servers just expose the raw value).

content/en/docs/Responses/AlbumID3WithSongs.md
```diff
@@ -161,11 +195,44 @@ description: >
...
 | `userRating` | `int` | No | **Yes**    | The user rating of the album. [1-5]|
+| `albumLabel` | `string` | No |  **Yes**   | The label producing the album. |
+| `musicBrainzId` | `string` | No |  **Yes**   | The album MusicBrainzID. |
+| `genres` | Array of `string` | No | **Yes**    | The list of all genres of the album. |
+| `artists` | Array of [`ArtistID3`](../artistid3) | No | **Yes**    | The list of all album artists of the album. |
+| `displayArtist` | `string` | No |  **Yes**   | The single value display artist. |
+| `releaseType` | `string` | No | **Yes**    | The type of release. If the server handle subtypes then the field should return them type and subtypes separated with " / ".|
+| `moods` | Array of `string` | No | **Yes**    | The list of all moods of the album. |
+| `sortName` | `string` | No |  **Yes**   | The album sort name. |
+| `originalReleaseDate` | `string` | No |   **Yes**   | Date the album was originaly released. [ISO 8601]|
```
epoupon, line 206: originally

content/en/docs/Responses/Child.md
Tolriq: Because this is a different value with a different meaning and it prevent any compatibility issue for server that in the V1 API decided to only return the first artist and ignore the others.

content/en/docs/Responses/ReplayGain.md
Tolriq: Well on my client the users requested and wanted a base gain for all then replay gains value added to it. Is there clients that offers a base replay gain only when there's no replay gain in tags?
Anyway no problem to be more clear about it, just wondering the usage in case I need to implement that in Symfonium.

content/en/docs/Responses/Child.md
sentriz: ok, so `artist` is the name of the entity that `artistId` points to, where as `displayArtist` is more true to what the user had in their tags

content/en/docs/Responses/ReplayGain.md
dweymouth: It's called fallback gain in this case and I want to add it to Supersonic eventually, but I will have to change my replayGain implementation a little bit. The usefulness of it is that most popular music tracks have a large negative gain when they are scanned, and if you have an unscanned track it will end up VERY loud relative to the others, unless you have a default fallback gain in place like -8 dB

content/en/docs/Responses/Child.md
Tolriq: Well it most importantly officially handle multiple artists. When original API only supported 1 and some servers did take only the first. The goal is to not touch legacy fields unless necessary to fix a bug.

content/en/docs/Responses/Child.md
```diff
@@ -11,30 +11,95 @@ description: >
...
+    "moods": [
+        "slow",
+        "cool"
+    ],
+    "replayGain": {
+        "trackGain": 0.1,
+        "albumGain": 1.1,
+        "trackPeak": 9.2,
+        "albumPeak": 9,
+        "baseGain": 0
```
dweymouth, line 101: What is the meaning of the baseGain property? It is not something from the ReplayGain standard

content/en/docs/Responses/ReplayGain.md
Tolriq: Hum no one ever asked that the base replay gain was enough. Will see if someone ask for that one day.
But if it's common should we not have something exposed by the server? Since the server can scan all the files and help finding a proper fallback gain no?

dweymouth: It would be something that should be configured in the client settings, IMO

dweymouth: image

ReplayGain settings from Supersonic. Eventually I may add a field for Fallback Gain eg. -8 db

content/en/docs/Responses/Child.md
Tolriq: It's given in the example https://datatracker.ietf.org/doc/html/rfc7845 for example, Opus can define that in the headers and it should be exposed to clients.

content/en/docs/Responses/ReplayGain.md
Tolriq: Yes so you do not have it yet ;) I have the same, but as said the server can scan all the files, get the peaks and expose a proper value. It's absurd to have it configured on all client side if it's always necessary it should be configured once on the server. I don't think that it have any drawback to anticipate and add a fallbackGain to the response.

dweymouth: I guess it's a difference in philosophy. I was thinking the fallback gain is a user preference, and since it's a playback detail should be configured on the client. But I suppose there's a case to be made for the server to take the average of all the replayGain values across the entire library and then expose that value as `fallbackGain` to the client, for files that are missing their own replayGain tags.

Tolriq: It is an user preference, but userRating and starred are user preferences too but we store them on the server so that every clients have them up to date and it support multiple users.
A fallbackGain here would also be user related since the queries are tied to the user and the server can either scan or offer a setting for the user to configure (Or I guess as 99% of the server will just ignore the field :p)
I'm adding it, even if no one implement it, it's still an useful reminder to clients that users may want that.

dweymouth: In that case I would think it would make sense to store all the user's ReplayGain preferences on the server, no? And a new API for `getUserReplayGainPreferences` that includes the preferred mode (track, album, or none), preamp, falback gain, and preventClipping

dweymouth: that's what I disagree with, fallbackGain being an immutable value tied to the songs - **unless** it's not a user setting but computed by the server as an average based on the rest of the library. In which case I'm happy for it to be a server-supplied value. But I think we're just talking past each other now and can wait for others to chime in :)

content/en/docs/Responses/AlbumID3.md
epoupon: For me this is an important field, it allows to filter Singles, EPs, Demos, Lives, etc. easily in the artist view.
Having a "free" string will make for sure each client parse for some values for each case it wants to support, like "EP", "ExtendedPlay", "Extended play" etc. (ignoring case)
But I guess it is better than not field at all.

content/en/docs/Responses/ReplayGain.md
Tolriq: Well this spam a little the others, but this is interesting discussion. I do not grasp why that value is not tied to the songs even if supplied as a user setting?
The goal is to have those songs not sound too loud when there's no gain, so it's by definition tied to all the songs and replay gain on some of the songs.
That difference of value between tagged songs and not tagged songs is something immutable, whatever the device that plays the media the difference in DB will be the same.

On the other end preAmp and clipping are personal choices that depends on where the sounds goes so does not fit a server side settings.

Or is there something else I'm missing here? Even with fallbackGain you'd still apply the preAmp (That could be negative to give head room)

content/en/docs/Responses/AlbumID3.md
Tolriq: In all cases single string or array, normalisation can't be enforced, user can put what they want in the tags.

From the logs I have from many users, there's tons of different values, and no easy way to normalize anything for those users.

In Symfonium I store all different values and exposes them to the smart filter stuff for easy selection, I do not care what the strings are and users are happy with that.

content/en/docs/Responses/ReplayGain.md
dweymouth: In a way you're right, even as a user setting it's the user's "best guess" as to what the average trackGain in their library is. But just as a user, IMO it would be confusing to put all my replaygain settings in the client, except fallback gain which I have to go configure on the server.

But as you mentioned the server knows all the replaygain values, so the server's "best guess" is going to be better than the user's best guess, which is why there's a case that the server should return fallback gain as the average of all the tracks that do have trackGain. And the server could do something even smarter if it wanted to, like if there is a single album track missing replayGain, the fallbackgain for that track could be set to the average for the rest of the album's tracks (or the album gain value for the album). Or it could be done by genre, because classical tracks with a missing replayGain value should probably be compared to other classical tracks with replaygain, rather than rock/pop. Or, the server could even run a replayGain scan using ffmpeg on files missing it, and just store actual replaygain values in its database and have no need for fallbackGain at all :)

All this is to say that the server is better at supplying a reasonable fallbackGain than the user would be, so if it's going to be a server value, it should be computed by the server IMO.

content/en/docs/Responses/AlbumID3WithSongs.md
```diff
@@ -161,11 +195,44 @@ description: >
...
+- `artists`
+- `displayArtist`
+- `releaseType`
+- `moods`
+- `sortName`
+- `originalReleaseDate`
+- `isCompilation`
+- `discTitles`
+
+**Note**: All OpenSubsonic added fields are **optionals**. But if a server support a field it **must** return it with an empty / default value when not present in it's database so that clients knows what the server supports.
```
epoupon, line 228: What do you mean by default value? Do you have examples?
If no disc title is available, empty array or array of empty strings?

Tolriq: Disc title is an object ;) But yes empty array in that case or empty string for the other cases.

If there's cases where default needs explicitness please comment on it.

content/en/docs/Responses/Child.md
```diff
@@ -11,30 +11,95 @@ description: >
...
+        },
+        {
+            "id": "ar-7",
+            "name": "Artist 7"
+        }
+    ],
+    "displayAlbumArtist": "Artist 6 & Artist 7",
+    "contributors": [
+        {
+            "role": "composer",
```
epoupon, line 70: what about "performer" with its subroles? (like drums, bass guitar, etc.) will see this later?

Tolriq: Never saw that on any provider, those are just roles like others no?
We can define a role subtype if necessary but it would simpler to maybe just define those as "performer/bass" role ?

content/en/docs/Responses/ReplayGain.md
kgarner7: I would say that the client should still have a way to set fallback gain (for example, a library with no replaygain tags whatsoever), but defaulting to the server computed value is probably fine (and how exactly it's done can be left to the individual servers). The _ideal_ (at least personally) would be server options to set the gain mode and preamp, but perhaps that's better as an enhancement later.

content/en/docs/Responses/AlbumID3WithSongs.md
epoupon: If 'isCompilation' (boolean) is supported but not known for this album, what are we supposed to put?

content/en/docs/Responses/Child.md
epoupon: This is yet another example of messy fields in tags (see https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#performer to see what picard does, according to file formats).
Again, this is still better than no field at all. But I mean, a server maintains the music library and should try hard to support a lot of format specificities and unify everything in its database (this is basically reconstructing the database used to tag the files). And the API should reflect that. If we just provide raw data that are read in tags and make all the server/client chain ignore their contents, I think we miss some presentation opportunities.
For example, the "credits" view of Roon: https://help.roonlabs.com/galleryDocuments/edbsn03b66b1efc4daf4a0b94d8ddaa04209a045cf8d85fb2d5879c34a0abb5510bf4d27ec0bf4c3a3c786f1e826116971ca8?inline=true
This is still possible with this proposal, but each client will have to further reparse role things, make guesses, etc.. At least, let's put an example with an instrument and/or chose a format that may simplify this.
(same for release types, we can present like this: https://help.roonlabs.com/galleryDocuments/edbsn6b5862c5f4c475b9b28dd151a3fa9f6100ecb1f9d69a41aed61af4666001672c17eb399b05eaa0555c095632270e3e5c?inline=true (see the albums/singles tabs), this is why I consider this field "important" and not just an info that users can build filters on)

content/en/docs/Responses/AlbumID3WithSongs.md
kgarner7: It would probably make sense to return false?

Tolriq: Yes it's an album by default like now.

content/en/docs/Responses/Child.md
Tolriq: The thing is that there's no magic on the servers, if the user have bad tags, the server can't normalize it, and if the user have proper tags then there's nothing to be done by the server. This is a tag software / user issue here more than anything. Not all servers use MB and not all users want to use MB anyway.

So we can put some recommendation in the docs to help users and servers that use MB data, but we can't enforce anything. Enforcing means that any data that does not fit the cases we expose can't be exposed and this breaks the liberty of the users to tag as they want to workaround some software limitations.

You have no idea how creative users are with the tags to sort their media via the wrongs tags because an app support that tag and not another.

With the roles you can already do the roon style thing, and the release value too.

So since you handle MB data that override tags, propose the format of the release and role field for subtypes and I'll add notes to the doc.

sentriz (dismissed):
lgtm, thanks for the effort on this 👌

also, do we need a new opensubsonic "capabilities" key for this?

content/en/docs/Responses/Child.md
epoupon: For roles I would say:
* "role", string
* "subroles", array of strings


For release types:
* "releaseType", string
* "secondaryReleaseTypes", array of strings

Tolriq: What I meant was a format in a single string :) The subparts are only relevant with MB are we sure we want to force that road to all?
Tags are supported by all servers, MB by nearly none so far.

epoupon: I understand your point, but mine is that if we want to aim to improve metadata quality, it would be better to introduce less open/clunky fields. Otherwise I am afraid they will be not that useful and likely difficult to modify afterwards.
But fine, let's just put a "role", that's better than nothing.

kgarner7 (approved):
My main concern was with ReplayGain (and other comments I had opinions on have been resolved), so LGTM

content/en/docs/Responses/Child.md
Tolriq: The problem is what do the server that only have 1 tag single string and need to fill those arrays that they can't? In all cases this will generate differences between servers.
I'm all in for more data, but we can't force servers to use MB. Most don't and won't, needs API keys, finding a working server, dealing with rate limits, ..... If there's an official way to tag the fields that servers can use to split the data and fill those arrays then I'm OK to add the arrays. If not we can only go with recommendation in a single string and then no arrays.
If the arrays are 100% specific to MB then I don't see the need and it's better to have a simple solution for you, a convention that role is performer/XXXX with XXX being the bass or whatever that client can use. Same for album release Album/XXXX for the sub types.
That covers your need + cover the users who already tag the media like that or they can edit their tags to match that. And this does not force the servers to handle MB while still allowing users to use the feature via the simple tags.

This covers all the cases without blocking people that would prefer to not support this than having to deal with tag normalisation and MB servers.

epoupon: Not sure why you consider this to be MB specific?
Not sure why "performer/XX,YY,ZZZ" would be better than another secondary field (except it brings yet again another questionable separator in a string). If a server does not have sub roles, then it just provides an empty array and that's all?
Not sure why it would block people to just put an empty array for this field?

As I said, the goal of all this is to make clients better. If no client is interesting in this, then this is fine, let's just put a "role", that's better than nothing.

Tolriq: You need to think from the other perspective.

At the start there's media files with tags.
releaseType and contributors are relatively well defined and are not multi valued.

So a user can only put a string in those fields and a server without MB can only read that string.

Now we have 2 possibility:
1) We expose on the API 1 string and we define a pattern that users can use.
This allows users and clients to properly work with that scheme without requiring the servers to try to normalize the data and be able to extract those sub values. (The only way would be that they do some separation on a specific separator, sounds familiar?)

2) We expose on the API 1 string and 1 array.
Servers now can either pass the data directly in the 1 string, or do some magic and pass it to the array sometimes, or the server have MB access and pass the data properly.

This means that on the client side, we still need to handle the 1 string + parsing of the well known cases + handling the array.
This brings absolutely no benefits to anyone in the end, just more works for the clients, when the only server that support MB can generate the string with the proper separator easily.

___

releaseType: "XXX" or "XXX/YYYY" or "XXXX/YYY/ZZZZ" 
Covers all needs if the clients support the separator he separates, first is type, rest are subtypes. Works via tags, via MB and without imposing anything impossible to do on the server side. We are normalizing the tags to align with MB it's a win win.

role: "XXXX" or "XXXX/YYYY" (There's never a case for more, the solution is to expose more roles)
Again this covers all the cases directly from tags, allows tags and clients to directly work together without forcing the server to do impossible things when it can't and still working with MB.

`{ "role" : "performer/Bass" , "artist": { .... }, "role" : "performer/leader" : "artist" : { ... } }`
Same artist can have multiple roles, using / as a separator is less an issue because the roles are known and there's none with a / it's not like we need to deal with AC/DC. 

So TL;DR; I do not think the arrays brings much and would only make the life of users more complicated to support advanced stuff if their server do not have MB access.

epoupon (approved):

sentriz:
re: https://github.com/tamland/airsonic-refix/pull/116#issuecomment-1739640169

before we ship, are we updating this PR to change genres to be [{name: "a"}] instead of ["a"] ?

content/en/docs/Responses/Child.md
kgarner7: Adding on one point of agreement, even if an artist is in multiple separate roles (for example, composer and vocals), would it not make sense for those to be separate entries? As far as I can tell (from Picard anyway), the role tags are `role:artist(s)`, so having multiple `{ "role": "...", "artist": "same" }` would reflect the tags as well

Tolriq:
Yes I will update later today or tomorrow morning my time, a little busy right now but it's almost done.

content/en/docs/Responses/Child.md
Tolriq: Yes they will be different entries, an artist can have multiple roles, but 1 role entry only contains one artist for simplicity, so if there's 3 composers there's 3 entries with role composer not 1 with 3 artists in it.
This simplify things and will better reflect answers when we add support for role filtering.

Tolriq:
Ok so pushed an update to the genres data that is now an array of ItemGenre to allow future addition of Genre ID. This is just a little more verbose, but will ensure we are not blocked in the future.

You need to review again sorry.

sentriz (dismissed):
:ok_hand:

epoupon (dismissed):

kgarner7 (approved):
:shipit:

content/en/docs/Responses/AlbumID3.md
```diff
@@ -11,19 +11,59 @@ description: >
...
+    "starred": "2023-03-22T01:51:06Z",
+    "duration": 1954,
+    "playCount": 97,
+    "genre": "Hip-Hop",
+    "created": "2023-03-10T02:19:35.784818075Z",
+    "artistId": "91c3901ac465b9efc439e4be4270c2b6",
+    "songCount": 8,
+    "played": "2023-03-28T00:45:13Z",
+    "userRating": 4,
+    "recordLabel": "Sony",
```
dweymouth, line 28: recordLabel should probably be an array too (for XML It'll just have to appear multiple times I guess) since albums can have multiple labels, especially for special editions

Tolriq: At the tag level no file format support a multi values field.

kgarner7: At least per [ID3v2.4 spec](https://github.com/id3/ID3v2.4/blob/516075e38ff648a6390e48aff490abed987d3199/id3v2.4.0-frames.txt#L253), there is formal definition of multi-valued tags (separated using `NULL` byte)

kgarner7: > All text information frames supports multiple strings, stored as a null
   separated list, where null is reperesented by the termination code
   for the charater encoding. All text frame identifiers begin with "T".

> TPUB
The 'Publisher' frame simply contains the name of the label or publisher.

At least from my reading there, this would mean that `TPUB` is also supported (I don't see any exceptions from the second definiton), as it is a text file. Mutagen and Taglib both handle this properly as well, although FFMpeg does not, so ymmv.

Tolriq: It also varies a lot on the tagger, TRCK, TIT2, TLEN and many other in theory also can be multi valued but no one would know what to do if there was multiple values. Anyway seems at least MP3tag support multi value for TPUB so let's change to array.

Tolriq:
And a new update pushed to address all the opened points including @epoupon one's.

kgarner7 (approved):
Sorry you keep having to do this, looks good!

Tolriq:
Well the normal plan is to do all those discussion in the discussion part, but seems people react only when it's too late ;)

content/en/docs/Responses/AlbumID3.md
```diff
@@ -59,12 +105,38 @@ description: >
 | `genre` | `string` | No |     | The album genre|
 | `played` | `string` | No | **Yes**    | Date the album was last played. [ISO 8601]|
 | `userRating` | `int` | No | **Yes**    | The user rating of the album. [1-5]|
+| `recordLabels` | Array of `string` | No |  **Yes**   | The labels producing the album. |
```
tamland, line 108: I guess recordLabels should follow the same pattern as genres? This will allow future list/filter api for record labels like genres.

content/en/docs/Responses/AlbumID3.md
```diff
@@ -59,12 +105,38 @@ description: >
...
+| `originalReleaseDate` | `string` | No |   **Yes**   | Date the album was originally released. [ISO 8601]|
+| `isCompilation` | `boolean` | No |  **Yes**    | True if the album is a compilation. |
+| `discTitles` | Array of [`DiscTitle`](../disctitle) | No | **Yes**    | The list of all disc titles of the album. |
 
 {{< alert color="warning" title="OpenSubsonic" >}}
 New fields are added:
 
 - `played`
 - `userRating`
+- `recordLabel`
```
tamland, line 125: Missing an `s`

content/en/docs/Responses/AlbumID3.md
Tolriq: Well at this point we can too, but will any server ever use that and expose ids? I'm not even sure any server will store more than 1 value.

Tolriq:
And a new update for recordLabels.

dweymouth (approved):

kgarner7 (approved):

epoupon (dismissed):
Thanks for this work!

Tolriq:
So @opensubsonic/clients @opensubsonic/servers I'll leave this opened for a week then it's merged and the start of real things :)

Some servers and clients are ready to go, but I prefer to leave some time in case someone came with something that could be a breaking change and be a pain later.

content/en/docs/Responses/ArtistID3.md
```diff
@@ -9,13 +11,20 @@ description: >
...
+    "id": "37ec820ca7193e17040c98f7da7c4b51",
+    "name": "2 Mello",
+    "coverArt": "ar-37ec820ca7193e17040c98f7da7c4b51_0",
+    "albumCount": 1,
+    "userRating": 5,
+    "artistImageUrl": "https://demo.org/image.jpg",
+    "starred": "2017-04-11T10:42:50.842Z",
+    "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
+    "sortName": "Mello (2)",
+    "roles": [
```
epoupon, line 23: Just thought: do we assume we skip subroles here?

Tolriq: Yes as said we don't want to force support on servers, so there probably won't be an API to filter on subroles.
Clients will be able to filter on artist have role "performer" then if they want to filter on subrole will filter at client side.

lachlan-00:
we are committing to NULL on supported? so i always give nothing even if its not available on the server? 
e.g the data  is dead but i care?

our stance is we support all or none? ampache will support all and empty on what is not availible like mood

Tolriq:
No null, some bad parser have issues with that and it's unclear about if it's null because you do not support the field or there's no value.

It's better to return empty strings; empty arrays and default values for the fields you support and not returning the field at all if not supported.

And no the stance is that you support what you want to support.

The only thing is that if you support a field you must return an empty value so that client knows that you support the field. If you do not support the field you do not return it.

So for Ampache you return all fields with empty values (not null) when no do data, and do not return the mood field at all.

content/en/docs/Responses/Child.md
```diff
@@ -100,11 +177,41 @@ description: >
 | `originalWidth` | `int` | No |     | The video original Width |
 | `originalHeight` | `int` | No |     | The video original Height |
 | `played` | `string` | No | **Yes**    | Date the album was last played. [ISO 8601]|
+| `bpm` | `int` | No |   **Yes**   | The BPM of the song. |
+| `comment` | `string` | No |  **Yes**    | The comment tag of the song. |
+| `sortName` | `string` | No |  **Yes**   | The song sort name. |
+| `musicBrainzId` | `string` | No |  **Yes**   | The track MusicBrainzID. |
```
epoupon, line 183: I just started implementing stuff so questions come just now...
Will we also need recording MBID? (used for example by listenbrainz for recommendations)

content/en/docs/Responses/AlbumID3.md
```diff
@@ -11,19 +11,66 @@ description: >
...
+            "id": "ar-2",
+            "name": "Artist 2"
+        }
+    ],
+    "displayArtist": "Artist 1 feat. Artist 2",
+    "releaseTypes": [
+        "Album",
+        "Remixes"
+    ],
+    "moods": [
```
epoupon, line 57: Not sure how these array of strings should be translated in XML.
Do you have an example?

sentriz: i'm no XML expert, but i've been using Go a guide for questions like this. in this case, the Go XML library would say something like 

(from https://go.dev/play/p/rf7X_q3XdUj)

```xml
<album>
<moods>dark</moods>
<moods>repetitive</moods>
</album>
```

content/en/docs/Responses/Child.md
Tolriq: This is the album MBid and it's exposed for the albums. I don't think there's a need to duplicate the data here since Subsonic have the scrobble endpoint. Else it's just a matter of requesting the album for the rare case it might be needed from a song no?

epoupon: Ah, you meant "release mbid" then. I would then suggest to rename the field to something more explicit, like `releaseMusicBrainzId`.
"track mbid" is used to identify a track within a release. "recording mbid" is used to identify a record, i.e some different tracks may actually point to the same recording (reeditions, compilations, etc.). Recording mbid is not only used for scrobling, but also to get recommendations. But all this is another story.

Tolriq: Well yes you are on child so song, so the mbID is the track MB ID. With that you can have access to everything from MB database.

content/en/docs/Responses/AlbumID3.md
```diff
@@ -59,12 +107,38 @@ description: >
 | `genre` | `string` | No |     | The album genre|
 | `played` | `string` | No | **Yes**    | Date the album was last played. [ISO 8601]|
 | `userRating` | `int` | No | **Yes**    | The user rating of the album. [1-5]|
+| `recordLabels` | Array of [`RecordLabel`](../recordlabel) | No |  **Yes**   | The labels producing the album. |
+| `musicBrainzId` | `string` | No |  **Yes**   | The album MusicBrainzID. |
```
epoupon, line 111: Technically this is the "release" mbid

content/en/docs/Responses/Child.md
epoupon: I don't query MB myself since everything I need is already in tags, but it looks like there is no easy way to query relationships using track MBID (would require listing tracks in the associated release? see "Lookups" in https://musicbrainz.org/doc/MusicBrainz_API + https://musicbrainz.org/doc/MusicBrainz_Entity)

Tolriq: This is a confusion between the tag names and the MB names.

epoupon: Yes indeed, that's why it would be better to make it clear in our specs (and IMO recording MBID is more useful than track MBID)
(by the way, Child is also used when browsing directories. In that case Child can be a directory that represents an album (with isDir=true), which is really messy for a generic mbid field. But that's another problem)

Tolriq: The question is more about what users tags and providers have, that what we think could be useful but no one use.
I base this work on what Kodi / Plex do and Kodi is currently by a very large margin the most advanced in this department.

The tags that they expect and report are:
- MUSICBRAINZ ARTIST ID
- MUSICBRAINZ ALBUM ARTIST ID
- MUSICBRAINZ ALBUM ID
- MUSICBRAINZ_TRACKID (Or for ID3 UFID with owner "http://musicbrainz.org" that is the track ID)

For the first 2 ones there's nothing to say.
For the trackID, as I said despite the name being track ID this is the recordingID
https://picard-docs.musicbrainz.org/downloads/MusicBrainz_Picard_Tag_Map.html

While we can enhance the doc later, the naming confusion was always here and users who use this are supposed to know, it's not really our job to document the tags.

For the album it's the release ID. Again we can update the docs later.

What is a lot more problematic is your remark about Child being used for other meanings. I completely missed that and it suxx because not only it can be album but it can also be artists.
And if we start to expose artistIds for songs we need to expose as arrays too and it start to get messy.

Else we need a way to differentiate the Child mediaType and expose a mediaType field that contains Track/Album/Artist so we can expose the corresponding MBId without a risk for client to not understand what kind of Id it refers to.

content/en/docs/Responses/AlbumID3.md
lachlan-00: @epoupon i installed subsonic 6.1.6 and recorded all my responses from the server here if you want xml examples

https://github.com/ampache/python3-ampache/tree/api6/docs/subsonic-6.1.6/xml-responses

content/en/docs/Responses/Child.md
epoupon: Or we could just always explicitely write: `artistMbid`, `releaseMbid`, `trackMbid`, `recordingMbid` and just set the relevant ones when needed (we could even set them all for a Child that is a song)
recording MBID encoded as MUSICBRAINZ_TRACKID and track MBID encoded as MUSICBRAINZ_RELEASETRACKID (or whatever) should be only an internal server problem. The server is here to handle various encoded tags from various formats.

Tolriq: I've just explained that it does not work a song have multiple artist so you can't return a single artistMbid we need to return arrays, but then arrays makes no sense when it's about an artist. And then we are missing the albumArtistMdIds that is also an array and should be returned too. But those can already be returned in the corresponding artists field on the songs so duplicate values for no gain.

Furthermore the basic rule of OpenSubsonic about those additional fields is that if they are supported they must be returned so a Child artist would have the other fields returned but empty, without anyway for the client to still be sure about what the Child is about anyway. So more confusion for no gain.

Then having trackMbid and recordingMbid / releaseMbid would bring back the naming confusion on our side. Since we expose values that does not correspond anymore to tag names and what all other providers do but the MB names that most people are not aware of at all.

Even the release Mb Id is named album Id in Picard.

Our goal is to fix Subsonic API not MB/Picard and the mess in the naming they build. Let's be pragmatic and consistent with what every other media providers do and avoid adding unnecessary confusion to something already confusing.

If the current proposal (With addition of the mediaType for child) does not work for you even if it works for all other providers then we remove the MB stuff from this and build another proposal just for that.

epoupon: I didn't say artistMbid is single valued. But indeed for a single Child artist this makes no sense to have an array.

I really don't understand why being explicit about mbid brings confusion, it is actually for me the exact opposite?

In picard, here is what we get in the UI:
image
Encoding these fields in messy/legacy named tags is really a problem that has nothing to do with our API.

Tolriq: It is and you are the confirmation that you are also affected :) You mix actual tags as what is in the files with MB names and what Picard shows you.

Your screen shot show **Tag** and the name **MusicBrainz Release Id** and yet this does not exist most other software will show another **Tag** name for that. The denomination of Tag in this case is an arbitrary internal name decided by Picard. (And in other software too).

And then there's the actual tags whose name even depend on the file and tag format. But for the MB Release ID they are all named as "Album Id". So there's no tag with the name Release Id never anywhere. 

This project is OpenSubsonic and deals with media files and tags, not what Picard hide behind the naming it wants. Many people have no idea Picard even exists.

So again either you agree on handling MB ids as the tags and how other media providers do and it's current proposal, else we remove them from this proposal and do it as another proposal. 
There's no need for endless discussion here about difference between tags and names in software that are not related to this one.

epoupon: I really don't care myself, internally I already use the "recording_mbid" and "track_mbid" denominations (as ListenBrainz requires recording_mbid or track_mbid json fields, not the things picard translates in tags).

That being said, if you all want trackMbid be the recording MBID (and possibly trackReleaseMbid be the track MBID), just because this is the name in the tags, that's perfectly fine for me, as long as we write in the comment what is is actually is.

Tolriq: Again this is not the proposal :) The exposed names are not trackMbId and whatever but **musicBrainzId**, same name for different entities that then map to what correspond the best in the internal naming of the tools or stuff you want to use.
We do not enter this endless naming debate on purpose because it's not our fight nor our concern.

The problem about our doc it that we will bring that debate again to expose all the case and tags and everything and document all the tags in each formats.

For the mapping I've already given https://picard-docs.musicbrainz.org/downloads/MusicBrainz_Picard_Tag_Map.html that shows all in a simpler way. But again this is a mapping that is internal and specific to Picard.

I use MP3Tag and https://docs.mp3tag.de/mapping/#musicbrainz_trackid the exposed name is MUSICBRAINZ_TRACKID and the concept of recording ID would mean absolutely nothing for the users of that software, it's the MB track ID for them.

And it's the same for many other tools. We do not care about the tools, we care about the tags, the tags have names that are easy to understand and commonly used. That's current proposal and it works for the other providers.

epoupon: Ok let's try to break this endless discussion.
* Are you happy with the current proposal or not?
* What is for you the musicBrainzId when Child is used for a song?

Tolriq: I already tried to stop the discussion twice ;) The current proposal works and is what others do. If you want something else we postpone the MB stuff. Each time you avoided the decision.

The Child will have a new mediaType field to improve the case when it's used for something else than songs (This helps for many other things than just MB so is beneficial by itself).

When Child is a song (so a track) this is the track ID that is in Picard / MB world the recording ID but in all the rest of the world the track ID.

If you remove Picard from the equation and take MP3Tag as the source then this discussion would not even exist as only TRACKID and ALBUMID exist and perfectly match the API.

If you remove all from the equation and only take the tags
We have track ID tag = the id of the track = MB recording ID and release track Id = the release track Id = MB release track ID.

The whole world and the tags match this definition, but Picard have decided otherwise and wants to show the MB names instead of the tag names to add some more confusion.

If you remove Picard a MB track ID is well know, and an MB album ID is well known and maps to recording / release in MB. 

Picard have decided that he wanted to hide the tags and do something else and confuse people, they can, but it's not our problem.

epoupon: You are focused on the names of the tags and on the tools/softwares that blindly read and repeat them. You already know that I don't agree with this approach. That's no big deal, let's end this discussion and let's go with your proposal!
But please put in the associated comment that in case of a song, the musicBrainzID is actually the recording MBID (and not the track MBID as it is stated now).

Tolriq: Again not my proposal, it's what all the others do and the names every one but Picard use. I offered you to build what you'd call a proper proposal but you do not seem to want to.

And I tried to explained I don't know how many times, that we are talking about software that read the tags and expose them ... Subsonic is a media provider that expose the media and their tags, they are not MusicBrainz clients that must use musicbrainz to have data to expose to clients.

I know my English is not that fluent but really maybe someone can jump in to explain...

99% of Subsonic servers that will expose the musicBrainzId will just blindly take the tag and expose it to the client.
For Songs they will take the tag that is named Track ID yes I know this is a stupid name but this is it's name not my call here just a fact ...

If the user actually put a release track ID there then 99% of the servers won't know, if a user put his mother name in that field then again 99% of the servers will never know.

So from there 2 solutions: We talk about tags and do not enforce anything, doc is correct. Else we enforce that the song musicBrainzId is really a recording ID and then it means that the server needs to validate the data and do calls to musicBrainz to ensure data integrity. Something I tried to explain would be a bad idea as we do not want to enforce such support on servers. Many servers would agree to blindly expose the tags, very few would add support for MB if they do not already support it. IMO having the data exposed without validation can still be a gain for some users and would be better than limiting the data to server that fully embrace MB. 

So yes this is an endless discussion, but this is an important one so that you understand the choices and agree with them, because that's the future of the API and servers that depends on those choices.

Anything enforced in the doc is enforcing, if we force tag data to match another website data, then servers must ensure data integrity and do the server calls.

In the same way we do not enforce anything in the sort name values, if a user want to put numbers in those tags to sort by a very own specific order it's his choice, and Subsonic will just blindly expose that data.

So TL;DR:
Do you understand why I talk about tags and not Picard / MusicBrainz?

And again: If you think only a proper MB integration is the way to go, then I remove this from this proposal and we build a proposal for MB support, something very different than just exposing a tag.

epoupon: Well I proposed trackMBID/recordingMBID/artistMBID/releaseMBID but you rejected it.
You want to stick to raw tags, fine. Let's go with your proposal, let's move on.

Tolriq: So you do not understand nor agree. Your half proposal is tied to MB not tags, I have no problem moving to that if that's what is preferred ....

I do not want to stick to anything I try to explain that we either use tags or MB and that's it's then a different API and a different work for the servers. But we can't be in the middle with half baked stuff. MB Ids are very important on the server side for a few things, but the usage on the client side is a lot more limited.

I don't know how to explain this more and you seems too angry to try anymore to read what I wrote.

This is a team work, even if I feel that if I do not push nothing moves, but it's important that everyone agree else this won't bring anything.

I'll remove the MB support from this proposal and we can discuss a better proposal when heads are cooled off. Let's not block this for something nearly no server currently support anyway.

epoupon: Don't get me wrong: your proposal works and I (really) do agree to move on it :)

What I propose is just to use the MusicBrainz names for our fields (as after all, we do expose their stuff, and that's how they refer to their own stuff in their docs + in their APIs) instead of sticking to the messy tag names.

For me, putting a `recordingMBID` field instead of `musicBrainzId` field does not imply the server needs to validate its content to the MB database, but yes indeed this implies all servers will need to be aware that MUSICBRAINZ_TRACKID is actually a recording MBID, even if they don't care (but I don't even expect them to check it is a properly formatted UUID).

But again, no problem to go the raw tags way, **really**.

sentriz: on the topic of using the proper MB name vs the tag name, is that not all subsonic servers are based around music tags. for example the gazelle-subsonic, and beets-subsonic projects. they serve a subsonic compatible API, but are not based on tags (they're based on the gazelle API and beets.io library, respectively. and they have explicit distinctions on release vs recording, release vs release group, etc)

Tolriq: The thing is that if they can do more they can do less :) The point is really about not enforcing things on servers that can't handle it while still allowing raw tag passing from servers to client that may want it.

In the same vein using the generic mbId name makes it non blocking for any future update that could expose all the MB data as the MB names are non clashing with this.

So let's go with this for now, I'll add the mediaType to child after the wait delay to avoid restarting again a round of votes.

There's still room to build something specific for MB the same way that subsonic API currently have hardcoded stuff for lastFM.

content/en/docs/Responses/AlbumID3.md
```diff
@@ -11,19 +11,66 @@ description: >
...
+    "playCount": 97,
+    "genre": "Hip-Hop",
+    "created": "2023-03-10T02:19:35.784818075Z",
+    "artistId": "91c3901ac465b9efc439e4be4270c2b6",
+    "songCount": 8,
+    "played": "2023-03-28T00:45:13Z",
+    "userRating": 4,
+    "recordLabels": [
+        {
+            "name": "Sony"
```
lachlan-00, line 30: Can we put id's into labels. They're also tracked with mbid's and the name can have many duplicates so name is too generic

Tolriq: The structure is prepared for future evolution, but there's no concept of recordLabel ids in Subsonic at all so it would require a major change to enforce.
If you talk about MB ids then there's no tag for those so require MB usage + server having the proper data.
This is not in the scope of this proposal and a new one can be made after.

content/en/docs/Responses/AlbumID3.md
```diff
@@ -11,19 +11,66 @@ description: >
...
+        "slow",
+        "cool"
+    ],
+    "sortName": "lagerfeuer (8-bit)",
+    "originalReleaseDate": "2001-03-10T02:19:35.784818075Z",
+    "isCompilation": false,
+    "discTitles": [
+        {
+            "disc": 0,
+            "title": "Disc 0 title"
```
epoupon, line 67: What if there is no disc title but just disc numbers?
Do you expect the discTitles array to be set with empty "title" fields?

Tolriq: The only OpenSubsonic rule is that a field must be returned if supported. So IMO the fact that you return an empty array for discTitles is enough for the client to know that you support it but that there's no discTitle for that album / disc.

I would only return empty title if the tag is explicitly empty. But even then I'm pretty sure any client would not display empty title names. (That's what I'll do when I'll add support for this one day).

epoupon: Ok, so just to be clear: if an album has a disc title for the first disc, but not for the second one (already hit this case in the past), `discTitles` with just contain the first disc?

Tolriq: I think both would be valid, whatever is simpler for the server.

content/en/docs/Responses/ReplayGain.md
```diff
@@ -0,0 +1,49 @@
...
+description: >
+  The replay gain data of a song.
+---
+
+{{< tabpane persistLang=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+    "trackGain": 0.1,
+    "albumGain": 1.1,
```
epoupon, line 15: I have encountered cases where the album gain was actually the disc gain (different gain for each disc of the same album)
Technically, this is no big deal here since ReplayGain is attached to a song, so "albumGain" would not be the same for all songs in the album.

Tolriq:
Ok guys so a week has passed, I've added the `mediaType` to Child and fixed the typos.

Requires again your review approval to finally merge this :)

And please review and approve https://github.com/opensubsonic/open-subsonic-api/pull/54 too.

tamland (approved):

sentriz (approved):
😎👍

epoupon (approved):

kgarner7 (approved):

Tolriq:
And it's merged :) We can now officially start implementing this on Clients and Servers and call OpenSubsonic real birthday today :)

PR #53, opened 2023-09-28
Document Stream endpoint behavior

Tolriq:
https://github.com/opensubsonic/open-subsonic-api/discussions/22 

Stream endpoint access must not count as playcount.

### DIFF
--- a/content/en/docs/Endpoints/stream.md
+++ b/content/en/docs/Endpoints/stream.md
@@ -1,6 +1,8 @@
 ---
 title: "stream"
-linkTitle: "stream"
+linkTitle: "stream [OS]"
+opensubsonic:
+- Clarification
 categories:
 - Media retrieval
 description: >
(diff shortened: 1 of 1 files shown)


Tolriq:
As discussed in https://github.com/opensubsonic/open-subsonic-api/discussions/22 (Top voted)

If you have not changed your mind, please add your reviews for merge.

sentriz (approved):
makes sense to me

ghenry22 (approved):
As per discussion, this is how navidrome already works and it’s a sensible change to have better control over playcount data quality.

anyone the records plays with navidrome will already work with this change.  Those that do not will simply have to add a call to the scrobble endpoint and remove any options client side to “scrobble to lastfm” as scrobbles will only go to the subsonic server.  Subsonic servers will have to implement an option to sync to lastfm or other external integrations.  Again Navidrome is a good example of this already well implemented in the server side.

epoupon (approved):

Tolriq:
So merging since it was debated earlier too and largely voted.

PR #58, opened 2023-10-10
Song Lyrics extension

Tolriq:
Add an extension to enable the songId parameter to getLyrics to ensure proper lyrics are returned for a given song (Embedded, local lrc, ...)

Since this is pretty basic I directly make the request here.

### DIFF
--- a/content/en/docs/Endpoints/getlyrics.md
+++ b/content/en/docs/Endpoints/getlyrics.md
@@ -17,10 +17,13 @@ Searches for and returns lyrics for a given song.
 | --- | --- | --- | --- | --- |
 | `artist` | No  |  |   | The artist name. |
 | `title` | No  |  |   | The song title. |
+| `songId` | No  | **Yes**  |   | The song ID. The [`Song Lyrics`](../../extensions/songlyrics/) extension enables this parameter. |
 
 ### Example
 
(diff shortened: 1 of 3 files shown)


kgarner7:
Adding on to the above, some things to consider:
- if you are going by tags, a file could have **both** synced and unsynced lyrics (or perhaps a server fetches from a third-party service). 
- When returning, you should give the ability to return both. This could either be something like `[{ "lyrics": "...", "synced": false }]` or `{ "synced": "...", "unsynced": "..." }`.
- it _might_ be nice to support asking for a specific format (but of course, that can't be guaranteed
- For synced lyrics, we should be explicit on the output lyric format. Is this just going to be a raw string (that the client needs to parse)? If so, what is the format? is it `[mm:ss.xx] line`? Or is `[mm:ss.xxx] line` allowed? Or do we expect the server to do the line splitting automatically like [Jellyfin](https://github.com/jellyfin/jellyfin/pull/8381)?

Tolriq:
This was a proposal to quickly fix the current API and allow clients to request explicit files.

If you want a proper lyrics extension then it's possible but needs a request and discussion in the discussion place.

Plex also have proper lyrics support for inspiration.

I'm out of energy for now to run after people for this, but if you reach a consensus I can make the actual extension request based on it.

PR #68, opened 2023-11-17
Add `releaseDate` field to AlbumID3

dweymouth:


Tolriq:
Maybe I was not clear enough in the discussion.

https://wiki.hydrogenaud.io/index.php?title=Tag_Mapping#Dates

The common current year stuff is recordingDate not releaseDate I thought this is what you wanted to expose (TRDC)

Maybe we should directly expose both?

dweymouth:
I am good with exposing both though `releaseDate` is the one I am most personally interested in. I'll update the PR to add `recordingDate` as well.

dweymouth:
Wait, actually does `recordingDate` only apply to tracks and not albums, as albums are often recorded over multiple dates, even multiple years?

Tolriq:
Honestly I don't know I don't use it @epoupon may know more here.

epoupon:
I only use picard, which only sets release level tags (date/year and originaldate/originalyear)

dweymouth:
Well my intiution then would be that `recordingDate` should only apply to tracks, and if we want to expose it, it could be a separate PR. What do you think @Tolriq

Tolriq:
I think that I hate MB ... They consider TDRC as the release date ... But TDRC is the recording date in the spec and other software. Should be TDRL .


They really like to do things their way that does not correspond to anything else.

dweymouth:
@opensubsonic/servers @opensubsonic/clients Marking this as ready to review under the assumption that `recordingDate` will be a property of tracks (`Child`) only and not album, and if we want that it will be a separate PR.

Tolriq:
@dweymouth we need to improve the doc, seeing the mess it is between the taggers.

Seems you want to go the Picard way, so the releaseDate needs to be made clear what tags it is TDRC & co and not TDRL as it might be supposed by a few.

dweymouth:
What I want to go with is that `releaseDate` is *the date the given issue in the user's library was released*. Which tags that comes from depends on the file type, no?

Tolriq:
Yes but the tags are not handled the same depending on the tagger that's the whole issue I'm trying to explain since a couple of messages :)

In ID3 V2.4 there's a tag named TDRC that is defined in ID3 standard as : RecordingTime, and there's a tag named TDRL that is ReleaseTime

Picard have decided that for it TDRC is the ReleaseDate this does not correspond to the tag standard.

Since we talk about releaseDate and recordingTime, we NEED to define what is what in the land of OpenSubsonic else it will be a mess server side too, as interpretation will vary and so the result.

I have no problem going the Picard way, but we need to be very explicit about that (and it will prevent us to handle recordingTime).

So to resume the mapping between releaseDate and the tags vary by tagger and not only by format, to avoid issue we need to take position too.

dweymouth:
We also might want to support the users who manually tag using fields like RELEASE_DATE as well, right? I think using semantic definitions of what the field `means` rather than what ID3v2 tag it's sourced from is what we want in a spec, right? And some servers may get this info from external info sources and not the embedded file tags at all

Tolriq:
The problem is not about all the other cases ...

The problem is about what are the servers supposed to do by default with what already exist in the tags.

They have a file with ID3 v2.4 and the field TDRC is filled.
The TAG spec says it's recording time and not release time, so they do not return the field because we define a meaning and we do not say that this tag is what we expect too? So the data you actually want by default would not be returned and you'd need to either retag to TDRL or ask the server to handle TDRC differently?

Open Subsonic is about removing the differences between the servers, so yes when we know there's a possible different handling of the TAG it's OpenSubsonic role to give a guidance about what we expect.

ID3 is the most used tag format, let's avoid future issues when they are already known no ?

dweymouth:
I think the best we can do is add language like "this usually comes from the TDRC field". We certainly don't want to say *only return this tag if it comes from the TDRC field* because that disallows servers from using external metadata sources. If there's a tagging difference between Picard and the TAG spec I guess we should go with the TAG spec, though the real definition of the field should still be meaning-based to allow for multiple sources.

Tolriq:
I know I'm not native English speaker, but well, I never said that we need to say it's only the TAG X ....

I said we need to clarify because the meaning of releaseDate depends on the tagger, and OS job is to help server in adding those fields, fields that by default most take from the tags.

I'm a server I see I need to add releaseDate it's the release date in the spec, ho nice, so let's see how it's in the tags. Ho there's multiple different variation, which one should I choose? Ho I'll take one in random since it's not defined.

Again OS goal is to try to avoid server differences.

So IMO we need to clarify. Either we decide it's the tag definition with or without a fallback to TDRC or we define it's TDRC if present.

"Due to difference between taggers and user interpretation for ID3 v2.4 releaseDate must/should be :"
- TDRC field if present else TDRL
- TDRL 
- TDRL field if present else TDRC.

This is not a big deal and will avoid future issues by being consistent. Don't know why every clarification always ends up with endless discussions.

dweymouth:
Because if we do that it's limiting what metadata sources are allowed. I think what we can do is define the field's meaning, say which ID3 field it should come from if present, but allow it to come from other sources as well. Not all file types even use ID3 tags also, though hopefully they use the same standard tag names. At some point, if the user's library is all mis-tagged and they get incorrect data, that's on them. We might have to wait for some servers to weigh in too. I am OK with saying "if this ID3 tag is present, the OpenSubsonic `releaseDate` field should come from it". I'm not OK with saying "the OpenSubsonic `releaseDate` field can *only* come from this ID3 tag".

Tolriq:
Read what I write maybe? Instead of just repeating arguments against something I never said nor implied? I'm writing exactly what you say, this is ridiculous.

Quote:
> I know I'm not native English speaker, but well, I never said that we need to say it's only the TAG X ....

I even give you a sample sentence that is exactly what you propose ....

> "Due to difference between taggers and user interpretation for ID3 v2.4 releaseDate must/should be :"
> - TDRC field if present else TDRL
> - TDRL
> - TDRL field if present else TDRC.

dweymouth:
I am not familiar with the ID3 tag names so I will let you decide which should take preference, but if we want to go with what you wrote, I'd amend it as such.

> Due to difference between taggers and user interpretation for ID3 v2.4 releaseDate must/should be :
>
 > *   TDRC field if present else TDRL, or if neither present any other appropriate metadata source
  > *  TDRL, or if not present any other appropriate metadata source
  > *  TDRL field if present else TDRC, or if neither present any other appropriate metadata source.

dweymouth:
BTW it looks like Navidrome is using the tag named `releasedate` (not TDRL or TDRC) to source this info: https://github.com/navidrome/navidrome/blob/a5dfd2d4a14ac5a5b87fc6016e57b94d2b964a69/scanner/metadata/metadata.go#L105

epoupon:
> I think that I hate MB ... They consider TDRC as the release date ... But TDRC is the recording date in the spec and other software. Should be TDRL .
> 
> They really like to do things their way that does not correspond to anything else.

They are actually aware of this:
"For compatibility reasons the date tag gets filled with the release date from MusicBrainz. This is how most software interprets this tag. Since Picard 2.9 the separate releasedate exists for use by scripts and plugins, but is not filled by default."

I remember some "endless" threads on their forums about dates...

IMO, we should expose releaseDate and originalReleaseDate at release level, and recordingDate at track level. From which tag to read this info should be a server only issue (with  settings server side, depending on how users tag their stuff)

dweymouth:
Agree. I think as clients we just have to trust that the info returned by the server is correct. And if it's not, it's a server or user library tagging issue, and there's nothing really the client can do about it.

Tolriq:
Again I don't care what and how it's written :) I'm not native speaker and won't write the best sentences.
I also personally don't care about the choice, so it's up to the servers if some already parse those tags or to clients who use those fields to vote for the standard.

All I want is that we take position about TDRC / TDRL (So picard vs TAG definition) because this is something servers have to decide and so we should provide them with an answer.

About Navidrome another misunderstanding of tagging ;) They use the value releasedate that is parsed by taglib and refactored depending on the tag format.

And as you can see: https://github.com/taglib/taglib/blob/master/taglib/mpeg/id3v2/id3v2frame.cpp#L316 
This is TDRL so they respect the tag standard , they parse TDRC as "date".

So we have Navidrome  using TDRL and LMS using TDRC (https://github.com/epoupon/lms/blob/master/src/libs/metadata/impl/TagLibParser.cpp#L339) (Date from Taglib that is TDRC)

Confirming the need to clarify .... so much energy lost.

Tolriq:
> we just have to trust that the info returned by the server is correct.

We are the servers too .... We need to define what is correct ....

dweymouth:
I think we're on the same page then @Tolriq - I'm ok with OpenSubsonic taking a position on the preference of TDRC vs TDRL, if we can get @deluan and @epoupon to agree. As long as we won't be *restricting* that the OpenSubsonic `releaseDate` must come from one or the other

Tolriq:
> As long as we won't be restricting

It was never suggested nor the plan. Just to document and decide on a known issue.

`originalReleaseDate` does not have this issue the tags are well defined without conflicts.

certuna:
> > I think that I hate MB ... They consider TDRC as the release date ... But TDRC is the recording date in the spec and other software. Should be TDRL .
> > They really like to do things their way that does not correspond to anything else.
>
> They are actually aware of this: "For compatibility reasons the date tag gets filled with the release date from MusicBrainz. This is how most software interprets this tag. Since Picard 2.9 the separate releasedate exists for use by scripts and plugins, but is not filled by default."

This was me :) A couple months ago I got fed up with Picard not supporting `releasedate` and I just submitted the PR to Picard to support `releasedate` in their internal tag mappings, so I could map it correctly myself with scripts. This essentially forced them to write this paragraph, they're well aware that they're doing it "wrong", but I also understand their position: there's too much software out there that ignores `TDRL`/releasedate, so a large part of their userbase doesn't use this tag.

So like with everything in tagging: if apps don't support something, people start doing workarounds. Apps don't support multivalued tags? People invent their own separators like semicolon. Apps don't support "Composer" (`TCOM`)? People put the composer in the Artist/Performer (`TPE2`) tag. Apps don't support Release Date (TDRL)? Just put it in Recording Date (TDRC).

> IMO, we should expose releaseDate and originalReleaseDate at release level, and recordingDate at track level. From which tag to read this info should be a server only issue (with settings server side, depending on how users tag their stuff)

Yep. I wrote the release date/original date/date mapping in Navidrome, and I'm using the "official specs", but I try to correct for Picard's way of tagging by remapping on import. It's...not particularly elegant but it works for "most" files out there.

Tolriq:
@dweymouth can you update the PR to the new format so that is done too?

Tolriq:
Since we debated can you add a small note about the fact that OS recommend TRDL as exposed by Taglib and that if a server uses another tag it should mention it in it's doc.

Tolriq (approved):

epoupon (approved):

kgarner7 (approved):

PR #80, opened 2023-12-22
Update originalReleaseDate field to new ItemDate type

dweymouth:
**NOTE: this is a breaking change** but as of now, LMS (server) and Symfonium (client) are the only users of this field, and they have coordinated the ability to make the change.

This is needed because the original type of the field was an ISO date which *does not* allow the representation of "partial" dates (ie only year known, or year and month), which is a common scenario with music file tags.

### DIFF
--- a/content/en/docs/Responses/AlbumID3.md
+++ b/content/en/docs/Responses/AlbumID3.md
@@ -59,7 +59,11 @@ description: >
         "cool"
     ],
     "sortName": "lagerfeuer (8-bit)",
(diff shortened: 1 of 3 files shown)


dweymouth:
CC @certuna

Tolriq:
The year needs to be optional too. If the server support the field it must return it not null, so in this case an empty object without year or month or day.

Tolriq (approved):

epoupon (approved):

deluan (approved):

PR #84, opened 2024-01-27
Add Nextcloud Music as a participating server

paulijar:
The Nextcloud Music app has a rudimentary support for the ~~OpenAmpache~~ OpenSubsonic API extensions starting from the version 1.10.0 (released today 2024-01-27).

The same application may be called also ownCloud Music, depending on the host cloud used.

### DIFF
--- a/content/en/docs/Endpoints/search3.md
+++ b/content/en/docs/Endpoints/search3.md
@@ -193,4 +193,5 @@ A [`subsonic-response`](../../responses/subsonic-response) with a nested [`searc
 | **Funkwhale** | 1.2.8 | Support not passing a query parameter. |
 | **Astiga** |  | Support `query=` |
 | **LMS** | 3.35.1 | Support `query=` |
+| **Nextcloud Music / ownCloud Music** | 1.10.0 | Support `query=` |
 {{< /alert >}}
--- a/content/en/docs/Responses/Subsonic-response.md
+++ b/content/en/docs/Responses/Subsonic-response.md
@@ -59,4 +59,5 @@ New fields are added:
 | **Funkwhale** |  |  Support `type`, version is exposed as `funkwhaleVersion` |
 | **Astiga** |  | Expose `Astiga/production` in `serverVersion` |
 | **LMS** |  | Support `type` |
+| **Nextcloud Music / ownCloud Music** | 1.10.0 | Support `openSubsonic`, `type`, and `serverVersion` |
 {{< /alert >}}
--- a/content/en/docs/_index.md
+++ b/content/en/docs/_index.md
@@ -33,6 +33,7 @@ Any server or client can join the organization and make proposals [OpenSubsonic]
 - [Gonic](https://github.com/sentriz/gonic)
 - [LMS - Lightweight Music Server](https://github.com/epoupon/lms)
 - [Navidrome](https://www.navidrome.org/)
+- [Nextcloud Music / ownCloud Music](https://github.com/owncloud/music)
 - [Supysonic](https://github.com/spl0k/supysonic)
 
 ### Clients


Tolriq:
OpenAmpache: it's OpenSubsonic :)

But nice! Do you plan to add support for extensions and additional fields? If yes do you want to be added to the OS org in the server group?

paulijar:
> OpenAmpache: it's OpenSubsonic :)

Oops, sorry, that's what I meant to say :D

> Do you plan to add support for extensions and additional fields? 

I probably will support at least some of them, eventually. Especially if there are any low-hanging-fruits.

> If yes do you want to be added to the OS org in the server group?

I'm not quite sure what are the implications of this. But if this doesn't come with any commitments, then it probably wouldn't hurt.

Tolriq:
> But if this doesn't come with any commitments

It does not really have any, it gives you a vote on new APIs. 

The only commitment here is not really tied to being in the org, it's just that if you propose an API, field, whatever and it's implemented you engage yourself in supporting it in your server or client as a sign of respect. But since you are the initiator it's mostly just common sense.

Tolriq (approved):

kgarner7 (approved):

sentriz (approved):

paulijar:
@Tolriq Okay, by all means, sign me in.

Tolriq:
@paulijar Seems there's some issues in the search3 end point returning duplicates : https://support.symfonium.app/t/no-tracks-listed-in-any-of-the-synced-albums/3512

You probably miss a sort in the query in that case most SQL servers with pagination results are undefined.

paulijar:
@Tolriq Thanks for the report. However, I couldn't immediately reproduce the problem. The DB queries here also sort the results by the title/name fields so that shouldn't be the issue here. What kind of API requests do you send when you see this?

Tolriq:
@paulijar The user is supposed to report to your repo I'll join there when he does it's not the good place :)

`https://xxxx/index.php/apps/music/subsonic/rest/search3.view?query=%22%22&musicFolderId=-1&songOffset=0&songCount=0&albumOffset=0&albumCount=500&artistOffset=0&artistCount=0&u=REDACTED&p=REDACTED&v=1.13.0&c=Symfonium&f=json`

But I think the root issue might be 
`https://xxxxxx/index.php/apps/music/subsonic/rest/getMusicFolders.view?u=REDACTED&p=REDACTED&v=1.13.0&c=Symfonium&f=json` that returns
```
{"subsonic-response":{"musicFolders":{"musicFolder":[{"id":-1,"name":"Artists"},{"id":-2,"name":"Folders"}]},"status":"ok","version":"1.16.1","type":"nextcloud music","serverVersion":"1.10.0","openSubsonic":true}}
```

And then both search3 with the 2 folderIds returns the same data.

Is this something normal?

paulijar:
Yes, that's normal. Is it a problem for your client? 

The thing is, Nextcloud Music abuses the folder system of the Subsonic API a bit. To understand why, it should be noted that the first reference clients I used 5 years ago were targeting quite old API versions and they didn't have any support for the metadata browsing using the methods like `getArtists`. Still, browsing by metadata is very much the core feature of Nextcloud Music so I came up with a solution where there's two fake folders: "Folders" and "Artists". The first of these presents the folders and files as they actually are stored in the file system. The second one creates an artificial directory hierarchy Artist > Album > Song according the metadata. 

There are never any other musicFolders except these two fake folders as we support only single top-level folder for the music library. This may be subject to change in the future, though. If that happens, then I need to rethink the use of musicFolders in the Subsonic API and I might decide to drop this legacy hack or at least make it configurable.

Tolriq:
The app sync all for offline first usage via search3. Since it sees 2 library it sync both so triggers duplicate that I can filter, but still duplicate all the requests and data so highly inefficient.

I can add a special handling to ignore the second one for your server type. Do you return different values for next cloud/owncloud ?

Are the IDs always the same? So that if the future you support real multiple folders you can avoid reusing the -2 to avoid me skipping a valid folder ?

paulijar:
The Subsonic API works exactly the same on Nextcloud and ownCloud. The IDs are always the same for these two fake folders: -1 and -2. If the real multi-folder support will come, then those folders will have positive ID values.

PR #94, opened 2024-06-30
Update support status for Nextcloud Music

paulijar:
Nextcloud Music v2.0.0 brought support for the endpoint `getLyricsBySongId`. For anyone wishing to test this, note that NC Music reads the lyrics only from the file metadata tags. The language information is not parsed but returned always as "xxx" in this API. The `offset` is returned always as "0" because NC Music already internally calculates the offset-adjusted timestamps and returns these for each line.

In addition to this new endpoint, NC Music v2.0.0 added a few simple properties defined in the OpenSubsonic API to a few responses.

Support for the endpoint `getOpenSubsonicExtensions` was added already in the previous release v1.11.0.

### DIFF
--- a/content/en/docs/Endpoints/getLyricsBySongId.md
+++ b/content/en/docs/Endpoints/getLyricsBySongId.md
@@ -120,3 +120,12 @@ Does not exist.
 | Field        | Type                          | Req.    | OpenS.  | Details                   |
 | ------------ | ----------------------------- | ------- | ------- | ------------------------- |
 | `lyricsList` | [`lyricsList`](../../responses/lyricslist) | **Yes** | **Yes** | List of structured lyrics |
+
+---
+
+{{< alert color="warning" title="OpenSubsonic server support" >}}
+
+| Server | Min vers. | Comment |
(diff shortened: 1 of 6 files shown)


Tolriq:
We have mostly stopped doing that server side and use a post in discussion.

It's too much work to maintain.

deluan:
Maybe we could add links to the pages each server/client dev maintains with their own compatibility list? This could be added in the [Overview](https://opensubsonic.netlify.app/docs/) page

paulijar:
If these compatibility lists are not to be maintained, then I think they should be removed to avoid confusion.

Tolriq:
That was the mid term plan, they were there at start to show that some servers did implement everything and motivate other servers and clients to join.

I think we now have reached a state where the project is known enough and the big players are here, so there's no more need to show off.

I agree with @deluan a page maintained by each project works well.

If you are motivated to remove the sections you can PR else I'll do that at some point, probably after the big holidays;

paulijar:
Okay, cool. I'll consider making the PR if I find the time and motivation, but no promises.

PR #109, opened 2024-10-24
Fix: Make getIndexes documentation consistent with Subsonic

kgarner7:
Currently, getIndexes only has `index` of type `IndexID3` (which has `ArtistID3`).
The original spec (per https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd) has `Indexes` with `shortcut` (`Artist[]`), `index` (`Index[]`) and `child` (`Child[]`), with `lastModified` and `ignoredArticles`.
Additionally, `Index` has an array of `Artist`, not `ArtistID3` (`getIndexes` organizes by not ID3). This does not change any behavior.

Inspired by discussion in https://github.com/navidrome/navidrome/pull/3422.
Also adds XML.

Finally, note that the `index` file is still named `indexid3`, because Hugo _really_ doesn't like just `index.md`.

Example data from https://www.subsonic.org/pages/inc/api/examples/indexes_example_1.xml


Tolriq:
Can we still rename to index_ or something to remove the id3 and avoid possible future confusion? Not that important.

Tolriq (approved):

epoupon (approved):

dweymouth (approved):

PR #119, opened 2024-12-07
Make old endpoints consistent with OG Subsonic, various bugfixes

kgarner7:
There are multiple endpoints currently that are an array that are required, which is not really aligned with [OG Subsonic spec](https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd).

Additionally, add some missing items (not videos, unless absolutely necessary), and a few minor fixes for other endpoints.


content/en/docs/Responses/jukeboxPlaylist.md
```diff
@@ -8,12 +8,83 @@ description: >
...
 
-| Field |  Type | Req. | OpenS. | Details |
-| --- | --- | --- | --- | --- |
+| Field          | Type                         | Req.    | OpenS. | Details                                      |
+| -------------- | ---------------------------- | ------- | ------ | -------------------------------------------- |
+| `currentIndex` | `int`                        | **Yes** |        | The current index of the song being played   |
+| `playing`      | `boolean`                    | **Yes** |        | Whether the queue is currently playing       |
+| `gain`         | `float`                      | **Yes** |        | Volume, in a range of [0.0, 1.0]             |
+| `position`     | `int`                        | No      |        | The current position of the track in seconds |
+| `child`        | Array of [`Child`](../child) | No      |        | The songs currently enqueued in the jukebox  |
```
jeffvli, line 90: This should be `entry` rather than `child`.

content/en/docs/Responses/PodcastEpisode.md
```diff
@@ -0,0 +1,74 @@
...
+
+{{< alert color="primary" title="OpenSubsonic" >}}
+`NowPlayingEntry` extends [`Child`](../child) (See [`Child`](../child) documentation for the corresponding fields)
+{{< /alert >}}
+
+| Field         | Type                              | Req.    | OpenS. | Details                                   |
+| ------------- | ----------------------------------| ------- | ------ | ----------------------------------------- |
+| `streamId`    | `string`                          | No      |        | ID used for streaming podcast             |
+| `channelId`   | `string`                          | **Yes** |        | ID of the podcast channel                 |
+| `description` | `boolean`                         | No      |        | Episode description                       |
```
jeffvli, line 72: `description` should be a `string` rather than boolean

content/en/docs/Responses/PodcastEpisode.md
```diff
@@ -0,0 +1,74 @@
...
+  "channelId": "17",
+  "description": "Jonas Gahr Støre fra Arbeiderpartiet er med i dagens partilederutspørring i Hallo P3!",
+  "status": "completed",
+  "publishDate": "2015-09-07T15:29:00.000Z"
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+{{< alert color="primary" title="OpenSubsonic" >}}
+`NowPlayingEntry` extends [`Child`](../child) (See [`Child`](../child) documentation for the corresponding fields)
```
jeffvli, line 65: Should be `PodcastEpisode` extends... ?

content/en/docs/Responses/PlayQueue.md
```diff
@@ -8,12 +8,86 @@ description: >
...
+  ]
+}
 {{< /tab >}}
 {{< /tabpane >}}
 
-| Field |  Type | Req. | OpenS. | Details |
-| --- | --- | --- | --- | --- |
+| Field       | Type                         | Req.    | OpenS. | Details                                             |
+| ----------- | ---------------------------- | ------- | ------ | --------------------------------------------------- |
+| `current`   | `int`                        | No      |        | ID of currently playing track                       |
```
jeffvli, line 88: `current` should be a `string` for the ID

content/en/docs/Responses/searchResult2.md
```diff
@@ -8,12 +8,135 @@ description: >
...
+}
 {{< /tab >}}
 {{< /tabpane >}}
 
-| Field |  Type | Req. | OpenS. | Details |
-| --- | --- | --- | --- | --- |
+| Field    | Type                           | Req. | OpenS. | Details         |
+| -------- | ------------------------------ | ---- | ------ | --------------- |
+| Field    | Type                           | Req. | OpenS. | Details         |
+| -------- | ------------------------------ | ---- | ------ | --------------- |
```
jeffvli, line 139: Extra table header here should be deleted

content/en/docs/Responses/user.md
```diff
@@ -8,12 +8,67 @@ description: >
...
+| `playlistRole`        | `boolean`      | **Yes** |        | Whether the user can create playlists        |
+| `coverArtRole`        | `boolean`      | **Yes** |        | Whether the user can get cover art           |
+| `commentRole`         | `boolean`      | **Yes** |        | Whether the user can create comments         |
+| `podcastRole`         | `boolean`      | **Yes** |        | Whether the user can create/refresh podcasts |
+| `streamRole`          | `boolean`      | **Yes** |        | Whether the user can stream                  |
+| `jukeboxRole`         | `boolean`      | **Yes** |        | Whether the user can control the jukebox     |
+| `shareRole`           | `boolean`      | **Yes** |        | Whether the user can create a stream         |
+| `videoConversionRole` | `boolean`      | **Yes** |        | Whether the user can convert videos          |
+| `avatarLastChanged`   | `string`       | No      |        | Last time the avatar was changed [ISO 8601]  |
+| `folder`              | Array of `int` | No      |        | Folder ID(s)                                 |
```
jeffvli, line 74: Should `folder` be using an array of `string` instead?

kgarner7: Nope. For some reason, in OG, folder IDs are **integers** instead of strings. That's the only one

content/en/docs/Responses/PlayQueue.md
kgarner7: This might be an original spec issue, but the `current` is defined as an int?
```xml
<xs:attribute name="current" type="xs:int" use="optional"/>
```

deluan: Yes, it is an issue with old spec. As Navidrome uses ID strings, I implemented this as string and it didn't break any OG clients.

paulijar: It seems that the OG Subsonic server actually does use an integer there so the spec is in line with the implementation: http://demo.subsonic.org/rest/getPlayQueue.view?c=test&v=1.16.1&f=json&u=guest&p=guest

But as that doesn't really make sense, should we still change it?

kgarner7: Wow. 
From that, it looks like internally all of the IDs are internally an int (included in a string), so it technically works. Given that it's inconsistent with the spec, I'd argue saying it should be a string is more correct? (I added an errata for now)

content/en/docs/opensubsonic-api.md
```diff
@@ -21,13 +21,13 @@ Correspondingly, there are two sets of methods for searching, starring and album
...
+| [Media retrieval](/categories/media-retrieval)               | [`stream`](../endpoints/stream) [`download`](../endpoints/download) [`hls`](../endpoints/hls) [`getCaptions`](../endpoints/getcaptions) [`getCoverArt`](../endpoints/getcoverart) [`getLyrics`](../endpoints/getlyrics) [`getAvatar`](../endpoints/getavatar) [getLyricsBySongId](../endpoints/getlyricsbysongid)                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
jeffvli, line 30: getLyricsBySongId is missing the tildes `, which makes it inconsistent with the rest of the endpoints

jeffvli (dismissed):

content/en/docs/Responses/videoInfo.md
```diff
@@ -17,3 +17,4 @@ description: >
 
 | Field |  Type | Req. | OpenS. | Details |
 | --- | --- | --- | --- | --- |
+z
```
lachlan-00, line 20: typo?

jeffvli (dismissed):

Tolriq:
I think there's question in the discussion part not resolved around the podcasts.
Since I don't use them I can't really comment.

jeffvli:
Given how few servers actually support podcasts, we should create a follow-up PR for that when discussion catches up. I'd prefer if we can get these initial docs fixes in first rather than be blocked by that.

deluan (dismissed):

jeffvli:
Fixed merge conflicts that resulted from #122, please re-review!

jeffvli (approved):

deluan (approved):

epoupon (approved):

PR #130, opened 2025-02-04
Add contributors to albumid3

kgarner7:
The definition of a contributor of an album can be server-specific (or just empty), but useful for lists.


dweymouth (approved):

Tolriq:
> The definition of a contributor of an album can be server-specific 

No it can't for OpenSubsonic since we want consistency on everything. A client needs to know what those contributors are to be able to make decisions.

lachlan-00:
this is pretty vague, it could be massive and full of information not really needed.

PR #135, opened 2025-03-16
Fix `artist` inconsistency

kgarner7:
- `artist` does not have `albumCount`. Fix the base type, and where this type is referenced
- the return value for `getArtist` is a specific type `artistwithalbumsid3`

See: https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd

### DIFF
--- a/content/en/docs/Responses/ArtistWithAlbumsID3.md
+++ b/content/en/docs/Responses/ArtistWithAlbumsID3.md
@@ -84,7 +84,7 @@ description: >
 
 | Field            | Type                               | Req.    | OpenS.  | Details                                                |
 | ---------------- | ---------------------------------- | ------- | ------- | ------------------------------------------------------ |
-| `id`             | `string`                           | **Yes** |         | The id of the arist                                    |
+| `id`             | `string`                           | **Yes** |         | The id of the artist                                   |
 | `name`           | `string`                           | **Yes** |         | The artist name.                                       |
 | `coverArt`       | `string`                           | No      |         | A covertArt id.                                        |
 | `artistImageUrl` | `string`                           | No      |         | An url to an external image source.                    |
@@ -93,7 +93,7 @@ description: >
 | `album`          | Array of [`AlbumID3`](../albumid3) | No      |         | Artist albums                                          |
 | `musicBrainzId`  | `string`                           | No      | **Yes** | The artist MusicBrainzID.                              |
 | `sortName`       | `string`                           | No      | **Yes** | The artist sort name.                                  |
-| `roles`          | Array of `string`                  | No      | **Yes** | The list of all roles this artist have in the library. |
+| `roles`          | Array of `string`                  | No      | **Yes** | The list of all roles this artist has in the library.  |
 
 {{< alert color="warning" title="OpenSubsonic" >}}
 New fields are added:
--- a/content/en/docs/Responses/ArtistWithAlbumsID3.md
+++ b/content/en/docs/Responses/ArtistWithAlbumsID3.md
@@ -1,0 +1,108 @@
+---
+title: "ArtistWithAlbumsID3"
+linkTitle: "ArtistWithAlbumsID3 [OS]"
+description: >
+  An extension of [`ArtistID3`](../artistid3) with [`AlbumID3`](../albumid3)
+---
+
+{{< tabpane persistLang=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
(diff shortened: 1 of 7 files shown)


content/en/docs/Responses/ArtistWithAlbumsID3.md
```diff
@@ -0,0 +1,108 @@
...
+      "starred": "2021-02-22T05:51:53Z"
+    }
+  ]
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field            | Type                               | Req.    | OpenS.  | Details                                                |
+| ---------------- | ---------------------------------- | ------- | ------- | ------------------------------------------------------ |
+| `id`             | `string`                           | **Yes** |         | The id of the arist                                    |
```
epoupon, line 87: Typo

content/en/docs/Responses/ArtistWithAlbumsID3.md
```diff
@@ -0,0 +1,108 @@
...
+| `name`           | `string`                           | **Yes** |         | The artist name.                                       |
+| `coverArt`       | `string`                           | No      |         | A covertArt id.                                        |
+| `artistImageUrl` | `string`                           | No      |         | An url to an external image source.                    |
+| `albumCount`     | `int`                              | **Yes** |         | Artist album count.                                    |
```
epoupon, line 91: It would be great to specify wether it is the number of albums this artist appears on or just the album this artist is credited as album artist

content/en/docs/Responses/ArtistWithAlbumsID3.md
```diff
@@ -0,0 +1,108 @@
...
+| `starred`        | `string`                           | No      |         | Date the artist was starred. [ISO 8601]                |
+| `album`          | Array of [`AlbumID3`](../albumid3) | No      |         | Artist albums                                          |
+| `musicBrainzId`  | `string`                           | No      | **Yes** | The artist MusicBrainzID.                              |
+| `sortName`       | `string`                           | No      | **Yes** | The artist sort name.                                  |
+| `roles`          | Array of `string`                  | No      | **Yes** | The list of all roles this artist have in the library. |
```
epoupon, line 96: This artist has

content/en/docs/Responses/ArtistWithAlbumsID3.md
kgarner7: That's a good question. Unfortunately I don't know how OG subsonic behaves.

deluan: IIRC, in OG subsonic `getArtist` only handles AlbumArtists, so this would be the number of albums this artist is credited as album artist.

epoupon: Maybe we could just say it has to be consistent with what the server returns when getting all the albums for this artist?
Because now with OS we likely want all albums (including albums where the artist "appears on") and this would be backward compatible.
Furthermore, poorly tagged albums don't have albumartist tags and servers usually fallback on artist in that case.
What do you think?

deluan: > Because now with OS we likely want all albums (including albums where the artist "appears on") and this would be backward compatible.

I'm not sure we should do this, as the client won't have a way to figure out what is the role of the artist in each album, as the AlbumID3 does not include the `contributors` attribute. Maybe we should add it?

kgarner7: I would also like to note that `getIndexes` (and `search2`) both return an `artist` (without album count). I get these are legacy endpoints, but I think it would might make more sense to only do `albumArtist`. Perhaps a new `getArtist2` endpoint that accepts a list of roles instead?

One other note, if this specific discussion goes on (or dies) without this specific question resolved, I don't think it should hold back this PR (as the point is to fix other endpoints).

deluan (approved):

dweymouth (approved):

content/en/docs/Responses/ArtistWithAlbumsID3.md
epoupon: > > Because now with OS we likely want all albums (including albums where the artist "appears on") and this would be backward compatible.
> 
> I'm not sure we should do this, as the client won't have a way to figure out what is the role of the artist in each album, as the AlbumID3 does not include the `contributors` attribute. Maybe we should add it?

Ah you are right, and not sure for contributors: I remember this pr https://github.com/opensubsonic/open-subsonic-api/pull/130 that was not approved. So indeed let's focus on one problem at a time

epoupon (approved):

PR #141, opened 2025-04-28
Fix data type of two NowPlayingEntry fields

khers:
The minutesAgo and playerId fields were marked as string and boolean respectively. In the example they are both integers.


kgarner7 (approved):

khers:
> Makes sense, and these fields are defined as integers also in the XML schema https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd.
> 
> Does anyone have any idea about the meaning and purpose of the `playerId` field? Does any client use it for anything?

In the original Subsonic, users could create multiple players to hold multiple play queues and transcoding rules. Mobile applications would also create players just for their use (DSub did this). I had a player for home and one for the office (mostly so play queues would stay separate).

lachlan-00 (approved):

PR #142, opened 2025-04-29
Add legacy album response and update search2Results

khers:
The page for search2Results has it returning a list of Child objects in the albums field but this is incorrect. The legacy API docs use an album type that is not in our documentation.

Add an entry for the legacy album type and update the search2 response type to use it.

### DIFF
--- a/content/en/docs/Responses/album.md
+++ b/content/en/docs/Responses/album.md
@@ -1,0 +1,268 @@
+---
+title: "Album"
+linkTitle: "Album"
+opensubsonic:
+- Change
+description: >
+  Album with optional songs.
+---
+
+{{< tabpane persistLang=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "id": "ad0f112b6dcf83de5e9cae85d07f0d35",
+  "name": "8-bit lagerfeuer",
+  "version": "Deluxe Edition",
+  "artist": "pornophonique",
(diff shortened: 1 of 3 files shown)


kgarner7 (changes requested):
No, per https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd `searchResult2` is defined as following

```xml
<xs:complexType name="SearchResult2">
<xs:sequence>
<xs:element name="artist" type="sub:Artist" minOccurs="0" maxOccurs="unbounded"/>
<xs:element name="album" type="sub:Child" minOccurs="0" maxOccurs="unbounded"/>
<xs:element name="song" type="sub:Child" minOccurs="0" maxOccurs="unbounded"/>
</xs:sequence>
</xs:complexType>
```

deluan:
I'm pretty sure the OG Subsonic XSD is correct, as it is used to drive all responses from the endpoints. Here's the search2 endpoint for reference: https://github.com/deluan/subsonic/blob/85c2eb6a731f65c6b3e39ab37dc858c12a75de0e/subsonic-main/src/main/java/net/sourceforge/subsonic/controller/RESTController.java#L777-L780

You can see it uses the JAXB library to prepare the responses, and the `Response` object is actually generated from the XSD.

`search2` is for folder-based browsing (as opposed to `search3`). If you check the docs and the XSD (and OG Subsonic codebase) you'll see that all Album references in the folder-based endpoints return a Child, not an AlbumID3

PR #144, opened 2025-05-09
separate playCount into totalCount and userCount

lachlan-00:
discussed here
https://github.com/opensubsonic/open-subsonic-api/discussions/96#discussioncomment-13076513

Essentially deprecate playCount for totalCount and userCount. If the servers don't use the same value for playCount might as well work towards removing it with 2 new items


epoupon:
Note that there are scenarios where privacy is important: we don't want a user to be able to deduce anything about other users, such as listening habits.
Therefore, I wouldn't expect every OS server to necessarily implement this field, or if it does, it should be configurable

paulijar:
For a reader not having prior knowledge of these discussions, it's not at all obvious that fields `totalCount` and `userCount` have anything to do with the times the song has been played. So maybe they should be `totalPlayCount` and `userPlayCount`, even though these require a bit more typing?

lachlan-00:
> Note that there are scenarios where privacy is important: we don't want a user to be able to deduce anything about other users, such as listening habits. Therefore, I wouldn't expect every OS server to necessarily implement this field, or if it does, it should be configurable

I would say that it's up to the server though to handle privacy but a good point. a count doesn't give you much information but it does give you information.

lachlan-00:
> For a reader not having prior knowledge of these discussions, it's not at all obvious that fields `totalCount` and `userCount` have anything to do with the times the song has been played. So maybe they should be `totalPlayCount` and `userPlayCount`, even though these require a bit more typing?

Sure I'll change that now.

PR #148, opened 2025-05-18
Add optional ISRC field to Child response

gabeklavans:
closes #123

### DIFF
--- a/content/en/docs/Responses/Child.md
+++ b/content/en/docs/Responses/Child.md
@@ -43,7 +43,12 @@ description: >
   "comment": "This is a song comment",
   "sortName": "Polar expedition",
   "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
-  "isrc": "USSM18300073",
+  "isrc": [
+    "USSM18300073",
+    "DELV42300297",
+    "DEE868300011",
+    "DEE868300007"
+  ],
   "genres": [
     {
       "name": "Hip-Hop"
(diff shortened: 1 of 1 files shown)


content/en/docs/Responses/Child.md
```diff
@@ -116,7 +117,6 @@ description: >
     "trackPeak": 9.2,
     "albumPeak": 9,
     "baseGain": 0
-  }
```
lachlan-00, line 119: typo? seems like that's needed

content/en/docs/Responses/Child.md
```diff
@@ -43,6 +43,7 @@ description: >
   "comment": "This is a song comment",
   "sortName": "Polar expedition",
   "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
+  "isrc": "USSM18300073"
```
lachlan-00, line 46: missing comma

content/en/docs/Responses/Child.md
gabeklavans: yup, typo

content/en/docs/Responses/Child.md
gabeklavans: clearly I need a nap

lachlan-00 (approved):

kgarner7 (changes requested):
Note that it is possible for one song (or recording) to have _multiple_ ISRCs.

gabeklavans:
> Note that it is possible for one song (or recording) to have _multiple_ ISRCs.

Very good point. Perhaps it's an array of strings, then, instead of a string?

gabeklavans:
> Though, is it possible for a music file to have multiple ISRCs in its metadata? It seems at least for ID3 that it should be one unique value for that file.

Sorry, edited my prior comment to add the above

kgarner7:
I don't see why not? For example, ID3v2.4 allows _all_ text frames to be multivalued (including `TSRC`, isrc). For an example of this in real life, see https://musicbrainz.org/recording/34f59f6d-9fdb-4bed-bdb8-06478011d9be

lachlan-00:
I didn't see multiples in my quick look but that should definitely be an array then

kgarner7 (dismissed):

kgarner7 (approved):

lachlan-00 (approved):

epoupon (approved):

PR #156, opened 2025-06-13
Add play queue extensions for using index

kgarner7:
The current `savePlayQueue` and `getPlayQueue` rely on a `current` being the id of an item in the queue. This is problematic if the queue has duplicate items, as it is then not possible to determine which index item to use.

This PR adds extensions `getPlayQueue2` and `savePlayQueue2` that rely on a 1-based `currentIndex` to clearly specify which item to use. Otherwise, they are identical to `getPlayQueue` and `savePlayQueue`.


content/en/docs/Endpoints/getplayqueue2.md
```diff
@@ -0,0 +1,83 @@
...
+linkTitle: "getPlayQueue2"
+categories:
+  - Bookmarks
+OpenSubsonic:
+  - Extension
+description: >
+  Returns the state of the play queue for this user.
+---
+
+`http://your-server/rest/getPlayQueue` Since [1.12.0](../../subsonic-versions)
```
lachlan-00: Using the old name?

content/en/docs/Endpoints/saveplayqueue2.md
```diff
@@ -0,0 +1,57 @@
...
+linkTitle: "savePlayQueue2 [OS]"
+categories:
+  - Bookmarks
+OpenSubsonic:
+  - Extension
+description: >
+  Saves the state of the play queue for this user.
+---
+
+`http://your-server/rest/savePlayQueue` Since [1.12.0](../../subsonic-versions)
```
lachlan-00: Same again. Old name

kgarner7:
> I like the number system.
> 
> It keeps things clear and tells me that they are similar functions with a major change or difference.
> 
> Calling the extension positionBased seems like that could be better.
> 
> playQueueByIndex or even just call the extension playQueue as we are expanding the playQueue functions

Thanks, I like this suggestion. I've changed the naming of the extension, endpoints and response type to be clearer.

Bar other asks, I'll add the OpenAPI next (once I figure out the necessary changes).

content/en/docs/Endpoints/getplayqueuebyindex.md
```diff
@@ -0,0 +1,83 @@
...
+  - Bookmarks
+OpenSubsonic:
+  - Extension
+description: >
+  Returns the state of the play queue for this user.
+---
+
+`http://your-server/rest/getPlayQueueByIndex` Since [1.12.0](../../subsonic-versions)
+
+Returns the state of the play queue for this user (as set by `savePlayQueue` or `savePlayQueueByIndex`). This includes the tracks in the play queue, the currently playing track index (1-based), and the position within this track. Typically used to allow a user to move between different clients/apps while retaining the same play queue (for instance when listening to an audio book).
```
dweymouth: I would suggest that this be 0-based because elsewhere in the API (e.g. `updatePlaylist`, there is already the concept of 0-based indexes)

lachlan-00: I'm a 0 fan, I think it's good after this.

lachlan-00 (approved):

content/en/docs/Endpoints/saveplayqueuebyindex.md
```diff
@@ -0,0 +1,57 @@
...
+{{< tab header="Subsonic" >}}
+Does not exist
+{{< /tab >}}
+{{< /tabpane >}}
+
+{{< alert color="warning" title="OpenSubsonic" >}}
+**Note** `id` is optional. Send a call without any parameters to clear the currently saved queue.
+In this case, `currentIndex` **must not** be set.
+
+If `currentIndex` is less than 1 or larger than the queue, the server **must** respond with error code 10.
```
lachlan-00: Sorry, found 1 more

openapi/endpoints/getPlayQueueByIndex/GetPlayQueueByIndexResponse.json
```diff
@@ -0,0 +1,20 @@
...
+                },
+                {
+                    "$ref": "../../schemas/SubsonicResponse/SubsonicFailureResponse.json"
+                }
+            ]
+        }
+    },
+    "externalDocs": {
+        "description": "GetPlayQueueByIndexSuccessResponse",
+        "url": "https://opensubsonic.netlify.app/docs/endpoints/getplayqueuebyindex/"
```
Tolriq: The link is wrong it should point to response/playqueuebyindex

Tolriq (approved):

paulijar:
Does adding the new endpoint `getPlayQueueByIndex` here *really* give us added value, as compared to just adding the `currentIndex` property to the existing endpoint `getPlayQueue`? The named extension is of course needed for the client to know that it can save the queue using the index. The endpoint `savePlayQueueByIndex` may be justifiable if we want to avoid adding new args to the existing endpoints (although I'm not certain I understand why this would be a problem here, when connected to a named extension).

paulijar:
> I want to avoid a situation where both `current` and `currentIndex` are set at the same time in `savePlayQueue`.

An alternative solution could be to document that this is forbidden and servers will return an error in such case. But I'm not strongly against the new endpoint `savePlayQueueByIndex`, either.

dweymouth (approved):

content/en/docs/Endpoints/saveplayqueuebyindex.md
lachlan-00: still on 1 here `is less than 1 or larger than the queue,`

lachlan-00:
I'm ready to approve once the text referencing the error is done.

i'm about to upload the functions in Ampache

lachlan-00:
> > I'm ready to approve once the text referencing the error is done.
> 
> Can you clarify? Is there something that needs to be changed?

```
If `currentIndex` is less than 1 or larger than the queue, the server **must** respond with error code 10.
```
This is the opposite of the docs that say 0-based

lachlan-00 (approved):

Tolriq:
@lachlan-00 why merge if there's still errors ? 
There's no benefits to quick merging, with luck no one notice, but if any one notice or any server is too quick and implement with the error this means more works for that person to report, then to PR then to review again and for the server to fix and the client that will fail on that server.

lachlan-00:
> @lachlan-00 why merge if there's still errors ? There's no benefits to quick merging, with luck no one notice, but if any one notice or any server is too quick and implement with the error this means more works for that person to report, then to PR then to review again and for the server to fix and the client that will fail on that server.

errors are resolved

PR #166, opened 2025-07-07
Add measurementScheme field to ReplayGain Response

khers:
The values presented by the ReplayGain object can be misinterpreted without knowing the standard used in calculating them. It is my understanding that while these values are all measured in decibles, the standards used for calculating them make ReplayGain incompatible with the ReplayGainV2 and EBUR128, but there may be a conversion rate or factor between ReplayGainV2 and EBUR128. However, I am still not entirely cleanr on the relationships between the three standards. However, a play back application could misinterpret the values leading to unexpected behavoir.

It is not clear to me if we should specify how a client should interpret the lack of this field or if that can be left up to individual clients or users.

See: https://github.com/opensubsonic/open-subsonic-api/discussions/165

### DIFF
--- a/content/en/docs/Responses/ReplayGain.md
+++ b/content/en/docs/Responses/ReplayGain.md
@@ -32,6 +32,7 @@ Does not exist.
 | `albumPeak` | `number` | No | **Yes**    | The album peak value. (Must be positive) |
 | `baseGain` | `number` | No | **Yes**    | The base gain value. (In Db) (Ogg Opus Output Gain for example) |
 | `fallbackGain` | `number` | No | **Yes**    | An optional fallback gain that clients should apply when the corresponding gain value is missing. (Can be computed from the tracks or exposed as an user setting.) |
+| `measurementScheme` | `string` | No | **Yes** | The name of the measurement scheme used to calculate gain values. Valid values are 'ReplayGain', 'ReplayGainV2', 'EBUR128'. |
 
 **Note**: If the data is not present the field must be ommited in the answer. (But the replayGain field on [`Child`](../child) must always be present)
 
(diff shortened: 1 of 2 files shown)


khers:
That works for RG2 and EBU standards, but not for RG1. The original ReplayGain (IIUC) was based on peak volume levels in a track rather than perceived loudness and if a system is build around RG2 or EBU it cannot use (and there is no conversion for) RG1 values. AFAICT there is no target LUFS for RG1.

khers:
> What is the data needed for and the purpose ? You can have RG2 data targeting EBU LUFS and R128 targeting RG2 LUFS.

Right, but if the values are ReplayGain V1 then it is targeting neither and this is the concern. If my understanding is incorrect I am happy to drop this PR entirely. I intend to use this to determine if I have compatible normalization data already present or not because a user of MusicAssistant said that their open subsonic server stores and reports ReplayGain values. The MusicAssistant maintainers settled on only supporting EBU R 128 for volume normalization, if the values stored are RG2 or EBU then we can use them, if they are RG1 then we cannot.

Tolriq:
Replay Gain is not really normalisation actually.

And I think you mix things here, there's 2 different concepts 
How the "volume" is analyzed (BS.1770 for RG 2 and EBU) and the target LUFS -18 and -23.
RG1 is rare those days but for most case is quite similar to RG2.

The EBU R128 require -23 and that's probably what they mean by only supporting that. But you can have RG2 targeting something else than -18 see REPLAYGAIN_REFERENCE_LOUDNESS tag.

And RG1 and RG2 actually use the same tags.

So really I think the important data is the target LUFS here that should allow you to do what you need.

khers:
I'll submit a new PR with this field instead.

PR #168, opened 2025-07-15
Add new transcoding extension

Tolriq:
New extension to have proper transcoding solution in OpenAPI.

This is a WIP to start the many discussions that this will bring.

### DIFF
--- a/content/en/docs/Endpoints/getTranscodeDecision.md
+++ b/content/en/docs/Endpoints/getTranscodeDecision.md
@@ -1,0 +1,141 @@
+---
+title: "getTranscodeDecision"
+linkTitle: "getTranscodeDecision"
+categories:
+- Transcoding
+description: >
+  Returns a transcode decision for a given media file.
+---
+
+`http://your-server/rest/getTranscodeDecision`
+
+Returns a transcode decision for a given media file. This endpoint is used by clients to determine if a media file should be transcoded before streaming. The server will analyze the provided client information and return a decision.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `songId` | **Yes** | | | The ID of the song to be transcoded. |
+| `offset` | No | | 0 | The time offset in seconds from which to start transcoding. |
+
+### Request Body
+
+The request body must be a JSON object containing the client's capabilities. See the [`ClientInfo`](../responses/clientinfo) documentation for more details.
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="ClientInfo" lang="json">}}
+{
+  "name": "Pixel 8 Pro",
+  "platform": "Sonos",
+  "maxAudioBitrate": 512000,
+  "maxTranscodingAudioBitrate": 256000,
+  "directPlayProfiles": [
+  {
+"container": "mp3",
+"audioCodec": "mp3",
+"protocol": "http",
+"maxAudioChannels": 2
+  },
+  {
+"container": "flac",
+"audioCodec": "flac",
+"protocol": "*",
+"maxAudioChannels": 2
+  }
+  ,
+  {
+"container": "mp4",
+"audioCodec": "flac,aac,alac",
+"protocol": "*",
+"maxAudioChannels": 2
+  }
+  ],
+  "transcodingProfiles": [
+  {
+"container": "mp3",
+"audioCodec": "mp3",
+"protocol": "http",
+"maxAudioChannels": 2
+  },
+  {
+"container": "flac",
+"audioCodec": "flac",
+"protocol": "*",
+"maxAudioChannels": 2
+  }
+  ],
+  "codecProfiles": [
+    {
+      "type": "AudioCodec",
+      "name": "mp3",
+      "limitations": [
+        { "name": "audioBitrate", "comparison": "LessThanEqual", "value": "320000", "required": true }
+      ]
+    },
+    {
+      "type": "AudioCodec",
+      "name": "flac",
+      "limitations": [
+        { "name": "audioSamplerate", "comparison": "LessThanEqual", "value": "192000", "required": false },
+        { "name": "audioChannels",  "comparison": "LessThanEqual", "value": "2",      "required": false }
+      ]
+    }
+  ]
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+### Example
+
+{{< alert color="primary" >}} `http://your-server/rest/getTranscodeDecision.view?songId=123&u=demo&p=demo&v=1.16.1&c=AwesomeClientName&f=json` {{< /alert >}}
+
+### Result
+
+A [`subsonic-response`](../../responses/subsonic-response) element with a nested [`transcodeDecision`](../../responses/transcodedecision) element on success.
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "subsonic-response": {
+    "status": "ok",
+    "version": "1.16.1",
+    "type": "AwesomeServerName",
+    "serverVersion": "0.1.3 (tag)",
+    "openSubsonic": true,
+    "transcodeDecision": {
+      "canDirectPlay": false,
+      "canTranscode": true,
+      "transcodeReason": ["AudioCodecNotSupported"],
+      "errorReason": "",
+      "transcodeUrl": "/rest/getTranscodeStream.view?trackID=123&offset=0&parameters=...",
+      "sourceStream": {
+        "protocol": "http",
+        "container": "flac",
+        "codec": "flac",
+        "audioChannels": 6,
+        "audioBitrate": 3000000,
+        "audioProfile": "",
+        "audioSamplerate": 96000,
+        "audioBitdepth": 24
+      },
+      "transcodeStream": {
+        "protocol": "hls",
+        "container": "mp4",
+        "codec": "aac",
+        "audioChannels": 2,
+        "audioBitrate": 256000,
+        "audioProfile": "xHE-AAC",
+        "audioSamplerate": 48000,
+        "audioBitdepth": 16
+      }
+    }
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `transcodeDecision` | [`transcodeDecision`](../../responses/transcodedecision)  | **Yes** |     | The transcode decision |
--- a/content/en/docs/Endpoints/getTranscodeStream.md
+++ b/content/en/docs/Endpoints/getTranscodeStream.md
@@ -1,0 +1,28 @@
+---
+title: "getTranscodeStream"
+linkTitle: "getTranscodeStream"
+categories:
+- Transcoding
+description: >
+  Returns a transcoded media stream.
+---
+
+`http://your-server/rest/getTranscodeStream`
+
+Returns a transcoded media stream. **Note:** Clients should not construct this URL manually. Instead, they should use the `transcodeUrl` provided in the response of the [`getTranscodeDecision`](../gettranscodedecision) endpoint.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `trackID` | **Yes** | | | The ID of the track to transcode. |
+| `offset` | No | | 0 | The time offset in seconds from which to start transcoding. |
+| `parameters` | **Yes** | | | Server-specific transcoding parameters, obtained from the `getTranscodeDecision` endpoint. |
+
+### Example
+
+{{< alert color="primary" >}} `http://your-server/rest/getTranscodeStream.view?trackID=123&offset=0&parameters=...&u=demo&p=demo&v=1.16.1&c=AwesomeClientName&f=json` {{< /alert >}}
+
+### Result
+
+The raw transcoded media stream. In case of an error, a standard HTTP error code is returned with a descriptive message.
--- a/content/en/docs/Responses/DirectPlayProfile.md
+++ b/content/en/docs/Responses/DirectPlayProfile.md
@@ -1,0 +1,27 @@
+---
+title: "DirectPlayProfile"
+linkTitle: "DirectPlayProfile [OS]"
+opensubsonic:
+- Extension
+description: >
+  Defines a direct play profile.
+---
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "container": "mp3",
+  "audioCodec": "mp3",
+  "protocol": "http",
+  "maxAudioChannels": 2
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `container` | string | **Yes** | **Yes** | The container format (e.g., mp3, flac, mp4). |
+| `audioCodec` | string | **Yes** | **Yes** | Comma-separated list of supported audio codecs. |
+| `protocol` | string | **Yes** | **Yes** | The streaming protocol. Can be `http` or `hls`. |
+| `maxAudioChannels` | integer | No | **Yes** | The maximum number of audio channels supported. |
--- a/content/en/docs/Responses/TranscodingProfile.md
+++ b/content/en/docs/Responses/TranscodingProfile.md
@@ -1,0 +1,27 @@
+---
+title: "TranscodingProfile"
+linkTitle: "TranscodingProfile [OS]"
+opensubsonic:
+- Extension
+description: >
+  Defines a transcoding profile.
+---
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "container": "mp3",
+  "audioCodec": "mp3",
+  "protocol": "http",
+  "maxAudioChannels": 2
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `container` | string | **Yes** | **Yes** | The container format (e.g., mp3, flac). |
+| `audioCodec` | string | **Yes** | **Yes** | The target audio codec for transcoding. |
+| `protocol` | string | **Yes** | **Yes** | The streaming protocol. Can be `http` or `hls`. |
+| `maxAudioChannels` | integer | No | **Yes** | The maximum number of audio channels for the transcoded stream. |
--- a/content/en/docs/Responses/transcodeDecisionResponse.md
+++ b/content/en/docs/Responses/transcodeDecisionResponse.md
@@ -1,0 +1,60 @@
+---
+title: "transcodeDecisionResponse"
+linkTitle: "transcodeDecisionResponse [OS]"
+opensubsonic:
+- Extension
+description: >
+  The response from the getTranscodeDecision endpoint.
+---
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "subsonic-response": {
+    "status": "ok",
+    "version": "1.16.1",
+    "type": "AwesomeServerName",
+    "serverVersion": "0.1.3 (tag)",
+    "openSubsonic": true,
+    "transcodeDecision": {
+      "canDirectPlay": false,
+      "canTranscode": true,
+      "transcodeReason": ["AudioCodecNotSupported"],
+      "errorReason": "",
+      "transcodeUrl": "/rest/getTranscodeStream.view?trackID=123&offset=0&parameters=...",
+      "sourceStream": {
+        "protocol": "http",
+        "container": "flac",
+        "codec": "flac",
+        "audioChannels": 6,
+        "audioBitrate": 3000000,
+        "audioProfile": "",
+        "audioSamplerate": 96000,
+        "audioBitdepth": 24
+      },
+      "transcodeStream": {
+        "protocol": "hls",
+        "container": "mp4",
+        "codec": "aac",
+        "audioChannels": 2,
+        "audioBitrate": 256000,
+        "audioProfile": "xHE-AAC",
+        "audioSamplerate": 48000,
+        "audioBitdepth": 16
+      }
+    }
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `canDirectPlay` | boolean | **Yes** | **Yes** | Whether the media can be played directly by the client. |
+| `canTranscode` | boolean | **Yes** | **Yes** | Whether the media can be transcoded by the server. |
+| `transcodeReason` | string[] | No | **Yes** | An array of reasons why transcoding is necessary. Possible values are: `AudioCodecNotSupported`, `AudioBitrateNotSupported`, `AudioChannelsNotSupported`, `AudioSampleRateNotSupported`, `ContainerNotSupported`, `ProtocolNotSupported`. |
+| `errorReason` | string | No | **Yes** | A description of an error that occurred. Empty string if no error. |
+| `transcodeUrl` | string | No | **Yes** | The URL to use for transcoding if `canTranscode` is true. See [`getTranscodeStream`](../endpoints/gettranscodestream). |
+| `sourceStream` | [StreamDetails](../responses/streamdetails) | No | **Yes** | Details about the source media stream. |
+| `transcodeStream` | [StreamDetails](../responses/streamdetails) | No | **Yes** | Details about the target transcoded stream if `canTranscode` is true. |
--- a/openapi/endpoints/getTranscodeStream.json
+++ b/openapi/endpoints/getTranscodeStream.json
@@ -1,0 +1,65 @@
+{
+  "get": {
+    "summary": "Get Transcoded Stream",
+    "description": "Returns a transcoded media stream. Clients should not construct this URL manually, but use the `transcodeUrl` returned by the `getTranscodeDecision` endpoint.",
+    "operationId": "getTranscodeStream",
+    "tags": [
+      "Transcoding"
+    ],
+    "parameters": [
+      {
+        "name": "trackID",
+        "in": "query",
+        "description": "The ID of the track to transcode.",
+        "required": true,
+        "schema": {
+          "type": "string"
+        }
+      },
+      {
+        "name": "offset",
+        "in": "query",
+        "description": "The time offset in seconds from which to start transcoding.",
+        "required": false,
+        "schema": {
+          "type": "integer",
+          "default": 0
+        }
+      },
+      {
+        "name": "parameters",
+        "in": "query",
+        "description": "Server-specific transcoding parameters. This should be obtained from the `getTranscodeDecision` endpoint.",
+        "required": true,
+        "schema": {
+          "type": "string"
+        }
+      }
+    ],
+    "responses": {
+      "200": {
+        "description": "The transcoded media stream.",
+        "content": {
+          "application/octet-stream": {
+            "schema": {
+              "type": "string",
+              "format": "binary"
+            }
+          }
+        }
+      },
+      "400": {
+        "description": "Bad Request - The request was malformed or invalid."
+      },
+      "401": {
+        "description": "Unauthorized - Authentication failed."
+      },
+      "404": {
+        "description": "Not Found - The requested track does not exist."
+      },
+      "500": {
+        "description": "Internal Server Error - The server encountered an error during transcoding."
+      }
+    }
+  }
+}
--- a/content/en/docs/Extensions/transcoding.md
+++ b/content/en/docs/Extensions/transcoding.md
@@ -1,0 +1,21 @@
+---
+title: "Transcoding"
+linkTitle: "Transcoding"
+OpenSubsonic:
+  - Extension
+description: >
+  Adds support for clients to make transcoding decisions and retrieve transcoded media streams.
+---
+
+**OpenSubsonic version**: [1](../../opensubsonic-versions)
+
+**OpenSubsonic extension name**: `transcoding` (As returned by [`getOpenSubsonicExtensions`](../../endpoints/getopensubsonicextensions))
+
+## Version 1
+
+This extension provides a way for clients to get a server's decision on whether a media file should be transcoded and to retrieve the transcoded stream if necessary.
+
+This extension requires the following endpoints:
+
+- [`getTranscodeDecision`](../../endpoints/gettranscodedecision): Returns a transcode decision for a given media file.
+- [`getTranscodeStream`](../../endpoints/gettranscodestream): Returns a transcoded media stream.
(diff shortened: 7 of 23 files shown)


content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,141 @@
...
+"container": "mp3",
+"audioCodec": "mp3",
+"protocol": "http",
+"maxAudioChannels": 2
+  },
+  {
+"container": "flac",
+"audioCodec": "flac",
+"protocol": "*",
+"maxAudioChannels": 2
```
kgarner7: In `DirectPlayProfile`, `protocol` is described as either `http` or `hls`. I understand the _intent_ of `*` here, but `*` is not specified as a valid value in the spec. If the intent is for it to be both `http` and `hls`, then I'd suggest either:
1. Make it an array
2. Codify `*` as allowing both methods

content/en/docs/Endpoints/getTranscodeStream.md
```diff
@@ -0,0 +1,28 @@
...
+
+Returns a transcoded media stream. **Note:** Clients should not construct this URL manually. Instead, they should use the `transcodeUrl` provided in the response of the [`getTranscodeDecision`](../gettranscodedecision) endpoint.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `trackID` | **Yes** | | | The ID of the track to transcode. |
+| `offset` | No | | 0 | The time offset in seconds from which to start transcoding. |
+| `parameters` | **Yes** | | | Server-specific transcoding parameters, obtained from the `getTranscodeDecision` endpoint. |
```
kgarner7: `parameters` is a very broad idea. Is this something that should be _repeated`? (e.g., `parameters=...&parameters=...`). I understand that the point is probably to allow servers to be as generic as possible, but it would be worth codifying _at least_ whether it's single-valued or multi-valued.

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `transcodeDecision` | [`transcodeDecision`](../../responses/transcodedecision)  | **Yes** |     | The transcode decision |
```
kgarner7, line 144: Either this url is wrong, or `content/en/docs/Responses/transcodeDecisionResponse.md` should be renamed

content/en/docs/Responses/transcodeDecisionResponse.md
```diff
@@ -0,0 +1,60 @@
...
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `canDirectPlay` | boolean | **Yes** | **Yes** | Whether the media can be played directly by the client. |
+| `canTranscode` | boolean | **Yes** | **Yes** | Whether the media can be transcoded by the server. |
+| `transcodeReason` | string[] | No | **Yes** | An array of reasons why transcoding is necessary. Possible values are: `AudioCodecNotSupported`, `AudioBitrateNotSupported`, `AudioChannelsNotSupported`, `AudioSampleRateNotSupported`, `ContainerNotSupported`, `ProtocolNotSupported`. |
+| `errorReason` | string | No | **Yes** | A description of an error that occurred. Empty string if no error. |
+| `transcodeUrl` | string | No | **Yes** | The URL to use for transcoding if `canTranscode` is true. See [`getTranscodeStream`](../endpoints/gettranscodestream). |
```
kgarner7: Rather than provide a url, maybe it'd be better to just expose it as `parameter` (if single-valued) or `parameters` (if multi-valued)

content/en/docs/Responses/transcodeDecisionResponse.md
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `canDirectPlay` | boolean | **Yes** | **Yes** | Whether the media can be played directly by the client. |
+| `canTranscode` | boolean | **Yes** | **Yes** | Whether the media can be transcoded by the server. |
+| `transcodeReason` | string[] | No | **Yes** | An array of reasons why transcoding is necessary. Possible values are: `AudioCodecNotSupported`, `AudioBitrateNotSupported`, `AudioChannelsNotSupported`, `AudioSampleRateNotSupported`, `ContainerNotSupported`, `ProtocolNotSupported`. |
```
kgarner7: Would it be worthwhile making this an enum instead?

content/en/docs/Responses/TranscodingProfile.md
```diff
@@ -0,0 +1,27 @@
...
+  "protocol": "http",
+  "maxAudioChannels": 2
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `container` | string | **Yes** | **Yes** | The container format (e.g., mp3, flac). |
+| `audioCodec` | string | **Yes** | **Yes** | The target audio codec for transcoding. |
```
kgarner7, line 25: Probably worthwhile being a bit more explicit here (`audio/*`, or just `codec`)?

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,141 @@
...
+      "type": "AudioCodec",
+      "name": "flac",
+      "limitations": [
+        { "name": "audioSamplerate", "comparison": "LessThanEqual", "value": "192000", "required": false },
+        { "name": "audioChannels",  "comparison": "LessThanEqual", "value": "2",      "required": false }
+      ]
+    }
+  ]
+}
+{{< /tab >}}
```
kgarner7, line 89: This block could be formatted a bit better

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,141 @@
...
+        "protocol": "http",
+        "container": "flac",
+        "codec": "flac",
+        "audioChannels": 6,
+        "audioBitrate": 3000000,
+        "audioProfile": "",
+        "audioSamplerate": 96000,
+        "audioBitdepth": 24
+      },
+      "transcodeStream": {
```
kgarner7, line 126: If a client specifies multiple transcoding profiles, and _multiple_ of then match, which one should the server use? The first? Last, Unspecified?

content/en/docs/Responses/DirectPlayProfile.md
```diff
@@ -0,0 +1,27 @@
...
+  "protocol": "http",
+  "maxAudioChannels": 2
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `container` | string | **Yes** | **Yes** | The container format (e.g., mp3, flac, mp4). |
+| `audioCodec` | string | **Yes** | **Yes** | Comma-separated list of supported audio codecs. |
```
kgarner7: Why a comma-separated list, instead of an array?

content/en/docs/Payloads/ClientInfo.md
```diff
@@ -0,0 +1,82 @@
...
+    {
+      "type": "AudioCodec",
+      "name": "flac",
+      "limitations": [
+        { "name": "audioSamplerate", "comparison": "LessThanEqual", "value": "192000", "required": false },
+        { "name": "audioChannels",  "comparison": "LessThanEqual", "value": "2",      "required": false }
+      ]
+    }
+  ]
+}
```
kgarner7, line 70: Formatting, again

content/en/docs/Payloads/Limitation.md
```diff
@@ -0,0 +1,27 @@
...
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `name` | string | **Yes** | | The name of the limitation. Can be `audioChannels`, `audioBitrate`, `audioProfile`, `audioSamplerate`, or `audioBitdepth`. |
+| `comparison` | string | **Yes** | | The comparison operator. Can be `Equals`, `NotEquals`, `LessThanEqual`, `GreaterThanEqual`, `EqualsAny`, or `NotEqualsAny`. |
+| `value` | string | **Yes** | | The value to compare against. For `EqualsAny` and `NotEqualsAny`, this should be a pipe-separated (`\|`) list of values (e.g.,  `44100\|48000`). |
+| `required` | boolean | **Yes** | | Whether this limitation must be met. |
```
kgarner7, line 27: What would be the point of a limitation that is not required?

content/en/docs/Payloads/Limitation.md
```diff
@@ -0,0 +1,27 @@
...
+  "comparison": "LessThanEqual",
+  "value": "320000",
+  "required": true
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `name` | string | **Yes** | | The name of the limitation. Can be `audioChannels`, `audioBitrate`, `audioProfile`, `audioSamplerate`, or `audioBitdepth`. |
```
kgarner7, line 24: Could this be an enum instead?

content/en/docs/Payloads/Limitation.md
```diff
@@ -0,0 +1,27 @@
...
+| `comparison` | string | **Yes** | | The comparison operator. Can be `Equals`, `NotEquals`, `LessThanEqual`, `GreaterThanEqual`, `EqualsAny`, or `NotEqualsAny`. |
+| `value` | string | **Yes** | | The value to compare against. For `EqualsAny` and `NotEqualsAny`, this should be a pipe-separated (`\|`) list of values (e.g.,  `44100\|48000`). |
```
kgarner7: Not too keen on having `equalsany` and `notequalsany` using `|` separated instead of a different datatype that has array).

kgarner7 (changes requested):
Only reviewed the markdown. General thoughts:
1. It's extremely important to describe the behavior of what happens when you specify multiple transcoding profiles. _Which_ one will the server return (it can only return one).
2. I'm a bit worried that the `limitations` would be a bit too heavy for a server. Not entirely sure.
3. Some parts (`codec`, `container`) could probably be made more explicit on the format

All things said, I do think this is a good first pass. I'm worried that it's a bit heavy for the server (especially if a large number of profiles/codecs/limitations is provided), but it should be good especially for mobile clients.

Tolriq:
>  It's extremely important to describe the behavior of what happens when you specify multiple transcoding profiles. Which one will the server return (it can only return one).

The transcoding profiles are in order of preferences, the server returns the first it can, the details are present  transcode decision answer.

> I'm a bit worried that the limitations would be a bit too heavy for a server. Not entirely sure.

Limitations are the core of the decision process, plex, emby, jellyfin all works with the same concepts, it's relatively easy to manage server side.

> Some parts (codec, container) could probably be made more explicit on the format

There was some discussion some times ago about returning details about the media and we did not reach a consensus.

I agree a consensus would be better as it would be used for the details returns for the tracks too, let's hope people can accept something.

content/en/docs/Endpoints/getTranscodeDecision.md
Tolriq: Yes * is valid, this is a WIP.

content/en/docs/Endpoints/getTranscodeStream.md
Tolriq: It should not be repeated, the server can encode what it want there.
Some servers may pass all ffmpeg parameters, some servers could just send a sessionID.
This was based on the fact that the endpoint returns the url.

content/en/docs/Responses/transcodeDecisionResponse.md
Tolriq: The problem I anticipated with that will be double encoding of the data by some clients if they handle url encoding differently than the server (+ vs %20, ...).
Not sure we can prevent all possible issues here.

content/en/docs/Responses/transcodeDecisionResponse.md
Tolriq: It would be in the OpenAPI but I don't think we ever defined enums in the API no ?

content/en/docs/Responses/TranscodingProfile.md
Tolriq: all the values are started with audioXXXX to anticipate an eventual support for music videos at some point.
It's codec like mp3,aac, ...

content/en/docs/Endpoints/getTranscodeDecision.md
Tolriq: The first it can, I thought it was written in the doc.

content/en/docs/Responses/DirectPlayProfile.md
Tolriq: Because the others do it like that and because of "*", but can be an array, just need to be clear about don't use * and something else.

content/en/docs/Payloads/Limitation.md
Tolriq: For optimizations clients side and directplay vs transcoding limitations.

Like the app can play 32b flac direct, but if you transcode then I'd like 16b flac.

content/en/docs/Payloads/Limitation.md
Tolriq: The problem is always users :) If we always put arrays, they can use the wrong comparison for multiple values.

And if we separate then either 2 different fields or unwanted polymorphism.

Tolriq:
This is not complicated ;)
This is necessary to have proper transcoding support for proper audio quality.

There's a lot of details in the discussion part.

Clients should have control on what quality they want. If I can't play some format like DSD, I do not want to receive low quality MP3, I want hi res FLAC.

When I cast to a Sonos device that does not support my FLAC 24/96 I want to receive FLAC 24/48 and now low quality mp3 or opus.

Same when casting to chromecast and a million other cases.

All major media providers have such an API and this is the main missing part of Subsonic for audiophiles and casting.

Tolriq:
@opensubsonic/servers So holidays are now mostly over :) Any objections or remarks on the draft would be nice to be done before the final polish and having to drop / rewrite everything.

sentriz:
also as a side, how do these changes interact with the `transcodedContentType` and `transcodedSuffix` which clients use?

lachlan-00:
> also as a side, how do these changes interact with the `transcodedContentType` and `transcodedSuffix` which clients use?

I've always treated these as the default output when running stream/download without additional options. User selectable format wouldn't affect the defaults outputs in that case.

content/en/docs/Endpoints/getTranscodeStream.md
```diff
@@ -0,0 +1,28 @@
...
+
+Returns a transcoded media stream. **Note:** Clients should not construct this URL manually. Instead, they should use the `transcodeUrl` provided in the response of the [`getTranscodeDecision`](../gettranscodedecision) endpoint.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `songId` | **Yes** | | | The ID of the track to transcode. |
+| `offset` | No | | 0 | The time offset in seconds from which to start transcoding. |
+| `paramtranscodeParamsters` | **Yes** | | | Server-specific transcoding parameters, obtained from the `getTranscodeDecision` endpoint. The value is unique and already properly escaped. |
```
epoupon: Just `transcodeParams` for the parameter

content/en/docs/Endpoints/getTranscodeStream.md
```diff
@@ -0,0 +1,28 @@
...
+linkTitle: "getTranscodeStream"
+categories:
+- Transcoding
+description: >
+  Returns a transcoded media stream.
+---
+
+`http://your-server/rest/getTranscodeStream`
+
+Returns a transcoded media stream. **Note:** Clients should not construct this URL manually. Instead, they should use the `transcodeUrl` provided in the response of the [`getTranscodeDecision`](../gettranscodedecision) endpoint.
```
epoupon: Since we now have `transcodeParams`, the note is no longer valid (the client does have to construct the URL manually)

content/en/docs/Responses/DirectPlayProfile.md
epoupon: I think we should then write down a note to say it is either "*" or a comma-separated list of audio codec names (and * is forbidden in the latter case)

content/en/docs/Payloads/Limitation.md
epoupon: Maybe I missed something: how do you specify this since a codec profile is not attached to neither a direct play profile nor a transcoding profile?

content/en/docs/Responses/transcodeDecision.md
```diff
@@ -0,0 +1,60 @@
...
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `canDirectPlay` | boolean | **Yes** | **Yes** | Whether the media can be played directly by the client. |
+| `canTranscode` | boolean | **Yes** | **Yes** | Whether the media can be transcoded by the server. |
+| `transcodeReason` | string[] | No | **Yes** | An array of reasons why transcoding is necessary. The strings are server specific made for logging purpose, the server should return 1 string per direct play profile. |
+| `errorReason` | string | No | **Yes** | A description of an error that occurred. Empty string if no error. |
```
epoupon: If the field is optional, no need to encode an empty string?

content/en/docs/Responses/TranscodingProfile.md
```diff
@@ -0,0 +1,27 @@
...
+  "audioCodec": "mp3",
+  "protocol": "http",
+  "maxAudioChannels": 2
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `container` | string | **Yes** | **Yes** | The container format (e.g., mp3, flac). Only a single value is accepted. |
```
epoupon: maybe we should say we don't expect/accept '*' here

content/en/docs/Responses/transcodeDecision.md
```diff
@@ -0,0 +1,60 @@
+---
+title: "transcodeDecisionResponse"
+linkTitle: "transcodeDecisionResponse [OS]"
+opensubsonic:
+- Extension
+description: >
+  The response from the getTranscodeDecision endpoint.
```
epoupon, line 7: Maybe we should also write a note about validity duration of this response.
Either we say it is server specific, either we add an optional validy duration in the response itself.

openapi/endpoints/getTranscodeStream.json
```diff
@@ -0,0 +1,65 @@
+{
+  "get": {
+    "summary": "Get Transcoded Stream",
+    "description": "Returns a transcoded media stream. Clients should not construct this URL manually, but use the `transcodeUrl` returned by the `getTranscodeDecision` endpoint.",
```
epoupon: outdated description

epoupon (approved):

lachlan-00:
How does a client send the clientinfo json body for getTranscodeDecision to the server?

This is the only request that requires the client send something outside parameters to the server as far as i know

lachlan-00:
Adding in a jsonPost extension let's you say something like

V1 - support json post body requests (required for getTranscodeDecision extension)

Then any future changes can be updated  there and i can also just support json body for all methods as well

Tolriq:
Why on earth so you want a second extension that have no purpose ?
The transcoding extension says the endpoint uses json post. Having another extension does not bring anything to the table.

What a client would do with that information ? The existing extension for post is necessary to indicate a change to old endpoint. 
A new extension would not indicate anything, no old endpoint would support those. So clients will just look at what extension they need and transcoding is auto sufficient as have all the needed information.

No json in that query also makes no sense, clients should use the old endpoint if they want default server config.

And that endpoint returns already an error if there's no posted json.

Again and extension for new feature is self contained, there's no need to add another extension that give no more information.

lachlan-00:
The purpose is you don't have it anywhere else in the api.

It's not in the original subsonic api so it's an extension

Tolriq:
Yes an extension named transcoding......

A jsonpost extension would add what information?
We have endpoints that may require jsonpost, but we can't tell witch one since the ones that will use it will require another extension.

The new extensions that needs json post have the information on them where it's useful.

A jsonpost extension gives absolutely 0 information about anything that is not already available in the proper other extensions.

Tolriq:
If you did not see it then the doc of the transcoding extension needs to be updated so you can see it. Adding something unrelated is not the solution.

Your proposal says need to support json post but not where so it serves no purpose as an extension.....

To have it as a proper extension it would require that v1 says. The end point gettranscode must support json.
When we add another on then we have v2, then v3 ....
This makes no sense as this brings 0 useful information, just duplicate things and force maintaining....

epoupon:
Hello!
@lachlan-00 Indeed, a new extension also seems overkill to me. If you missed it while reading the PR, it seems to me that if we clearly add that the client info must be in the body of the POST request for getTranscodingDecision, that does the job?

lachlan-00:
It just needs to be more visible in some way, It's semi-hidden not clear and will be missed by other people.

I'll not debate what is or isn't too much but it's not enough for me.

Tolriq:
The thing is that I'm not you so I can't guess "some way". Can you try to explain what more details / sentence you need in the end point ?

Currently it says in big.

> Request Body
The request body must be a JSON object containing the client’s capabilities. 

As me, Request Body in big in the header carries all the necessary information, so I really don't know what more may make this more clear.

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,146 @@
...
+
+Returns a transcode decision for a given media file. This endpoint is used by clients to determine if a media file should be transcoded before streaming. The server will analyze the provided client information and return a decision.
+This endpoint must be accessed using a POST request
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `mediaId` | **Yes** | | | The ID of the media to be transcoded. |
+| `mediaType` | **Yes** | | | Either song or podcast so the server knows what the mediaId is refering to. |
```
epoupon: "referring to"

epoupon (dismissed):

lachlan-00:
The change is significant enough that it needs to be somewhere other than a line in `Endpoints/getTranscodeDecision.md` pointing to a payloads folder that didn't exist before.

Both the server and the client need to do something that isn't done in OS or Subsonic and it's not explicit enough.

I don't accept "I don't want to" as a valid reason for not providing information in the documentation.

Put it in the transcode extension/other areas, expand on it, whatever you think but do something that gives the reader clarity about what this is.

This would have been approved and resolved weeks ago with small effort. I'm not going to keep repeating myself and will mute the thread.

epoupon (dismissed):

dweymouth (dismissed):

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,146 @@
...
+    "version": "1.16.1",
+    "type": "AwesomeServerName",
+    "serverVersion": "0.1.3 (tag)",
+    "openSubsonic": true,
+    "transcodeDecision": {
+      "canDirectPlay": false,
+      "canTranscode": true,
+      "transcodeReason": ["AudioCodecNotSupported"],
+      "errorReason": "",
+      "transcodeParams": "0001-0005-004",,
```
paulijar: An extra comma here

content/en/docs/Endpoints/getTranscodeStream.md
```diff
@@ -0,0 +1,28 @@
...
+
+Returns a transcoded media stream. **Note:** Clients should not try to reconstruct the `transcodeParams`. Instead, they must use the `transcodeParams` provided in the response of the [`getTranscodeDecision`](../gettranscodedecision) endpoint.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `songId` | **Yes** | | | The ID of the track to transcode. |
+| `offset` | No | | 0 | The time offset in seconds from which to start transcoding. |
+| `transcodeParams` | **Yes** | | | Server-specific transcoding parameters, obtained from the `getTranscodeDecision` endpoint. The value is unique and already properly escaped. |
```
paulijar: It took me a while to realize that this doesn't mean that the client should pass the entire result of `getTranscodeDecission` (which would be silly) but there is property `transcodeParams` within that result. Maybe the description could be extended a little bit: "Server-specific transcoding parameters, obtained from the property `transcodeParams` of the result of the endpoint `getTranscodeDecision`".

Tolriq: For the record it's also written above with 

> **Note:** Clients should not try to reconstruct the `transcodeParams`. Instead, they must use the `transcodeParams` provided in the response of the [`getTranscodeDecision`](../gettranscodedecision) endpoint.

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,146 @@
...
+"audioCodec": "mp3",
+"protocol": "http",
+"maxAudioChannels": 2
+  },
+  {
+"container": "flac",
+"audioCodec": "flac",
+"protocol": "*",
+"maxAudioChannels": 2
+  }
```
kgarner7: This should be indented properly like the rest of the body.

content/en/docs/Payloads/DirectPlayProfile.md
```diff
@@ -0,0 +1,27 @@
...
+  "maxAudioChannels": 2
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `container` | string | **Yes** | **Yes** | Comma-separated list of supported container format (e.g., mp3, flac, mp4) or `*` to not limit to a specific container. |
+| `audioCodec` | string | **Yes** | **Yes** | Comma-separated list of supported audio codecs or `*` to not limit to specific codecs. |
+| `protocol` | string | **Yes** | **Yes** | The streaming protocol. Can be `http`, `hls` or `*`. |
```
kgarner7: I get that it's implied, but probably be better to explicitly state that `*` means both.

content/en/docs/Payloads/Limitation.md
kgarner7: I still disagree that you should use pipe-separated. If you have some fields being an array, then for those comparisons only, just have a `valueArray`

content/en/docs/Payloads/ClientInfo.md
```diff
@@ -0,0 +1,82 @@
...
+"container": "mp3",
+"audioCodec": "mp3",
+"protocol": "http",
+"maxAudioChannels": 2
+  },
+  {
+"container": "flac",
+"audioCodec": "flac",
+"protocol": "*",
+"maxAudioChannels": 2
```
kgarner7: Same thing here, please indent to match formatting of the rest of the body.

content/en/docs/Endpoints/getTranscodeStream.md
```diff
@@ -0,0 +1,28 @@
...
+| `offset` | No | | 0 | The time offset in seconds from which to start transcoding. |
+| `transcodeParams` | **Yes** | | | Server-specific transcoding parameters, obtained from the property `transcodeParams` of the result of the endpoint `getTranscodeDecision`. The value is unique and already properly escaped. |
+
+### Example
+
+{{< alert color="primary" >}} `http://your-server/rest/getTranscodeStream.view?songId=123&offset=0&parameters=...&u=demo&p=demo&v=1.16.1&c=AwesomeClientName&f=json` {{< /alert >}}
+
+### Result
+
+The raw transcoded media stream. In case of an error, a standard HTTP error code is returned with a descriptive message.
```
kgarner7, line 31: For `/stream.md`, it's explicitly stated that this is an XML document (the standard error). Is this still the case here?

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,146 @@
+---
+title: "getTranscodeDecision"
+linkTitle: "getTranscodeDecision"
+categories:
+- Transcoding
```
kgarner7, line 5: This is also an OpenSubsonic `Extension`

content/en/docs/Endpoints/getTranscodeStream.md
```diff
@@ -0,0 +1,28 @@
+---
+title: "getTranscodeStream"
+linkTitle: "getTranscodeStream"
+categories:
+- Transcoding
```
kgarner7, line 5: This is also an OpenSubsonic `Extension`

content/en/docs/Payloads/DirectPlayProfile.md
kgarner7: If this is intended to be multivalued, why not just make it an array?

kgarner7 (changes requested):

Tolriq:
@kgarner7 Again "If you have some fields being an array, then for those comparisons only, just have a valueArray" is a no, I will not add polymorphism to the API with different fields being used based on different other fields in the same object.

It would be highly confusing to always use an array for an equal comparison that can only handle 1 value (And error prone)
And it would be insanely more complicated for servers / and or API wise to have to deal with multiple different limitations for the same codecs.

I can't think of another solution that is clean and usable. If you have something better then propose, but no to polymorphism there's no reasons to cause problems to many clients and auto generated API wrappers.

Tolriq:
And actually this also applies to the comma separated codecs and the * if we follow your logic against parsing of the strings.

If you prefer I can use comma separated values to match the rest of the proposal, else I'll ask all others to vote on the split versus changing everything.

kgarner7:
You can already have conflicting limitations by just specifying `<= number` and `> number` for the same field. If you absolutely want to eliminate the chance of conflicting limitations, then you would need to do something like the following, where each field is an optional _single_ rule.
```json
{
    "audioChannels": { "comparison": "LessThan", "value": 20 },
    "audioBitrate": {},
    "audioProfile": {},
    "audioSamplerate": {},
    "audioBitdepth": {}
}
```
Of course, if you _want_ to support multiple separate limitations for the same value (e.g., one that is required and one that is not), then this alternate schema would prevent that.

kgarner7:
When all is in the same list, you still have problems of incompatible rules, because the same item can be specified multiple times regardless. At the end of the day, you are trying to encode a list for certain parameters. In JSON, there's a type for that: it's called an array. 

At this point, I don't really care about the comparisons either way. I'd love for `=` and `!=` to be formally expressed as an array in the schema, but /shrug/. That being said, at least for the other types that are _always_ a list, you should make it so. And then, rather than say `*`, just make the list empty, since as I understand, the point of the `container`, `audioCodec` and `protocol` lists is to _restrict_ the allowable fields.

epoupon (approved):
Thanks a lot for all this work!

content/en/docs/Endpoints/getTranscodeDecision.md
```diff
@@ -0,0 +1,148 @@
...
+      "audioCodecs": [ "flac", "aac", "alac" ],
+      "protocols": [],
+      "maxAudioChannels": 2
+    }
+    ],
+    "transcodingProfiles": [
+    {
+      "container": "mp3",
+      "audioCodec": "mp3",
+      "protocol: "http",
```
paulijar: There is a closing quotation mark missing here, causing the rest of the example to be highlighted as an error on the web UI.

paulijar (approved):

epoupon (approved):

Tolriq:
No need to remove if you want to use that for your webinterface, but we can keep the API as it is to avoid forcing others to support that use case that was out the original idea.

PR #218, opened 2026-03-05
feat: songLyrics extension version 2 — word/syllable-level timing

ranokay:
## Summary

This PR adds **version 2** of the `songLyrics` extension to the OpenSubsonic spec and OpenAPI docs.

Version 2 introduces optional word/syllable-level karaoke timing behind `enhanced=true`, while keeping the default response fully backward compatible with version 1.

It also documents richer lyric modeling for sources such as TTML:

- independent `kind` tracks (`main`, `translation`, `pronunciation`)
- word/syllable timing via `cueLine` + `cue`
- shared per-track vocal attribution via `structuredLyrics.agents`
- `cueLine.agentId` references instead of repeating singer/layer metadata on every cueLine

This PR follows up on discussion #213 and incorporates the review feedback on cue ordering, overlap rules, and multi-agent attribution.

### New response types

- **`cue`**: a single timed word or syllable with `start`, optional `end`, and `value`
- **`cueLine`**: a timed line-level grouping of `cue` items
- **`agent`**: reusable per-track attribution metadata for cueLines

### New/extended fields

- **`structuredLyrics.kind`**: classifies each lyric track as `main`, `translation`, or `pronunciation`
- **`structuredLyrics.cueLine`**: word/syllable-level timing data parallel to `line`
- **`structuredLyrics.agents`**: optional per-track attribution metadata for cue-attributed lyrics
- **`cueLine.agentId`**: references an `agent` in the same `structuredLyrics` entry

### Endpoint changes

- **`getLyricsBySongId`** gets an **`enhanced`** boolean parameter on both GET and POST
- when `enhanced=true`, the response may include:
  - `kind`
  - `cueLine`
  - `agents`
  - non-main lyric tracks such as `translation` and `pronunciation`
- when `enhanced=false` or omitted, the response remains version 1-compatible

## Contract clarifications added during review

- `cueLine` is only meaningful for `synced=true`; unsynced lyrics must not include it
- within a single `cueLine`, `cue.end` is **all-or-none**
- when the source has partial cue end-times, servers **must** fill the missing ones
- cues inside a single `cueLine` **must not overlap**
- overlaps across different `cueLine`s are still valid, since those represent parallel vocal layers
- `agents` are scoped to a single `structuredLyrics` entry
- `agents[].id` must be unique within that entry
- if `agents` are used for cue-attributed lyrics, there must be exactly one `role: "main"` agent
- when multiple `cueLine`s share the same `index`, the one whose agent has `role: "main"` must come first
- if `agents` is present, every `cueLine` in that entry must include `agentId`
- each `agentId` must resolve to one local `agents[].id` in the same `structuredLyrics` entry
- `structuredLyrics` entries are independent across all `kind` values, including `main`
- clients must not assume 1:1 alignment of `line` arrays or `cueLine` arrays between `main`, `translation`, and `pronunciation`
- cue counts may differ across tracks for the same lyric passage

## Examples added/updated

- enhanced Korean example with full second-line `cueLine` coverage
- pronunciation example showing different cue counts from the main track
- background-vocals example using `agents` + `agentId`
- multiple-singers example using shared `agents` instead of repeated per-line role/name fields

## Backward compatibility

- `enhanced` defaults to `false`
- without `enhanced=true`:
  - the response shape is identical to version 1
  - no `cueLine` arrays are returned
  - no `agents` arrays are returned
  - non-main `kind` tracks are not returned
  - the existing `line` array remains unchanged

Servers supporting this version should advertise:

- `songLyrics` versions `[1, 2]`

### New

- `content/en/docs/Responses/agent.md`
- `content/en/docs/Responses/cue.md`
- `content/en/docs/Responses/cueLine.md`
- `openapi/schemas/Agent.json`
- `openapi/schemas/Cue.json`
- `openapi/schemas/CueLine.json`

### Updated

- `content/en/docs/Endpoints/getLyricsBySongId.md`
- `content/en/docs/Extensions/songLyrics.md`
- `content/en/docs/Responses/structuredLyrics.md`
- `openapi/endpoints/getLyricsBySongId.json`
- `openapi/openapi.json`
- `openapi/schemas/StructuredLyrics.json`

Closes #213


Tolriq (approved):

Tolriq (dismissed):

content/en/docs/Responses/cueLine.md
```diff
@@ -0,0 +1,56 @@
...
+| `value` | `string`                      | No      | **Yes** | Full text of the line (for validation/fallback)                                                                                                                                                                                                                                                                                                                                 |
+| `role`  | `string`                      | No      | **Yes** | Semantic role of the cues in this cueLine. Omitted or empty for main vocals. Values: `bg` (background vocals), `voice1` through `voiceN` (individual voice parts — N is any positive integer with no upper bound, e.g. `voice1`, `voice100`, `voice1000`), `group` (group/chorus vocals). When multiple cueLines share the same `index`, the main vocals cueLine **must** come first |
```
kgarner7: I'm not sure I agree with a specific required format for `role`: for example, if you have multiple vocals at the same time, and it's credited for say, named characters/roles (instead of just `voice1`, etc).

content/en/docs/Endpoints/getLyricsBySongId.md
```diff
@@ -118,6 +121,196 @@ Does not exist.
...
+- `cueLine` is a **parallel** structure, not a replacement for `line`
+
+Servers that don't support TTML or word-level timing simply never include these fields. Clients that don't support karaoke display simply ignore them.
+{{< /alert >}}
+
+{{< alert color="primary" title="cueLine behavior" >}}
+
+- `cueLine` data is only meaningful when `synced=true`. Servers **must not** emit `cueLine` arrays for unsynced lyrics.
+- Within a `cueLine`, `cue.end` **must** be either present on **all** cues or **none** (all-or-nothing). When the source provides partial end times, servers **must** fill missing values. When no cues have end times, `end` is omitted from all cues.
+- When multiple cueLines share the same `index`, the main vocals cueLine (empty/omitted `role`) **must** come first.
```
kgarner7: Is it guaranteed that the main lyrics will be tagged in such a way to distinguish that it is main? If not, then this expectation is not necessarily realistic.

content/en/docs/Endpoints/getLyricsBySongId.md
```diff
@@ -118,6 +121,196 @@ Does not exist.
...
+- Source data may contain overlapping cue intervals. Servers **should** normalize overlaps where possible. Clients should handle overlapping intervals gracefully.
```
kgarner7: What does normalizing overlaps mean?

content/en/docs/Endpoints/getLyricsBySongId.md
```diff
@@ -118,6 +121,196 @@ Does not exist.
...
+- Cues where `start == end` (zero-duration) may occur. Clients should treat these as instantaneous markers.
+- A pronunciation `structuredLyrics` entry may have a different number of cues per line than the main entry. Clients should not assume 1:1 cue correspondence between `kind` tracks.
+- A translation `structuredLyrics` entry may have fewer `line` entries than the main entry. Clients should not assume line arrays are the same length across `kind` tracks.
```
kgarner7: What about the behavior for kind `main`?

content/en/docs/Endpoints/getLyricsBySongId.md
```diff
@@ -118,6 +121,196 @@ Does not exist.
...
+          "cueLine": [
+            {
+              "index": 0,
+              "start": 2747,
+              "end": 6214,
+              "cue": [
+                { "start": 2747, "end": 3179, "value": "nuneul" },
+                { "start": 3582, "end": 4100, "value": "tteun" },
+                { "start": 4500, "end": 6214, "value": "sungan" }
+              ]
```
kgarner7, line 208: Is it intentional that the `main` `cueLine` and `pronunciation` `cueLine` don't include the second `line`? I know it would make the example longer, but it feels like it might be more useful to include that.

kgarner7 (changes requested):

content/en/docs/Responses/cueLine.md
Tolriq: This is named role because it's the role and is necessary for display purpose to be well defined. What you want is a mapping between role and names that would be another data on the lyric.

ranokay: The intent here is a normalized presentation role, not raw performer/character metadata. I clarified the docs so `role` stays a well-defined vocal-layer field for client rendering, while named singer/character mapping remains out of scope for this v2 proposal.

content/en/docs/Endpoints/getLyricsBySongId.md
ranokay: I adjusted the wording so this is conditional: if a default/main `cueLine` exists for a given `index`, it is represented by an empty/omitted `role` and must come first. Clients should not assume every source can distinguish a main layer.

content/en/docs/Endpoints/getLyricsBySongId.md
ranokay: I replaced the vague ‘normalize overlaps’ wording with concrete behavior around derived end-times and client handling, so the contract is explicit instead of hand-wavy.

content/en/docs/Endpoints/getLyricsBySongId.md
ranokay: I clarified that tracks are independent across all `kind`s, including `main`. `main` identifies the primary lyric layer for that entry, but clients should not assume 1:1 line or cue alignment across tracks.

content/en/docs/Endpoints/getLyricsBySongId.md
ranokay: Agreed. I expanded the example so the second `line` also has corresponding `cueLine` data where applicable.

openapi/schemas/CueLine.json
```diff
@@ -0,0 +1,41 @@
...
+		"index": {
+			"type": "integer",
+			"description": "Zero-based index into the parent line array this cueLine corresponds to"
+		},
+		"start": {
+			"type": "integer",
+			"description": "Start time in milliseconds (may differ from the parent line if cues are more precise)"
+		},
+		"end": {
+			"type": "integer",
```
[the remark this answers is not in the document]
ranokay: I don't want to take this in this PR. The inconsistency with Line timing types is noted, but changing the new v2 cue timing fields from integer to number should be a separate repo-wide schema decision rather than a drive-by change here. Milliseconds are naturally integers.

openapi/schemas/Cue.json
```diff
@@ -0,0 +1,26 @@
...
+		"description": "Cue",
+		"url": "https://opensubsonic.netlify.app/docs/responses/cue/"
+	},
+	"properties": {
+		"start": {
+			"type": "integer",
+			"description": "Start time in milliseconds"
+		},
+		"end": {
+			"type": "integer",
```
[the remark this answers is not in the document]
ranokay: Same reasoning here: I don't want to widen the new v2 timing fields in this PR without making a deliberate decision about timing types across the whole OpenAPI surface.

content/en/docs/Responses/cueLine.md
kgarner7: I still think this can do better:
1. Rather than have `voiceN`, if there are some types types that are numbered, it would make more sense to separate it out into `role` (an enum) and `index` (a 0-based field, denoting the index of the voice
2. You have a chance to do it right. Add an optional `displayRole`, or something similar.

content/en/docs/Endpoints/getLyricsBySongId.md
kgarner7: Thanks! Can you also update `content/en/docs/Responses/structuredLyrics.md` to have the cueLine reflect the same changes you did here?

content/en/docs/Endpoints/getLyricsBySongId.md
kgarner7: Thanks. Is there a specific reason why servers _should_ avoid it (and not **must** for example)?

ranokay: Good point — since this rule covers derived (server-generated) end-times rather than existing source timing, there's no reason a server can't guarantee non-overlapping values it's creating itself. Promoted to `must`. Clients still handle overlapping intervals from source data gracefully regardless.

content/en/docs/Endpoints/getLyricsBySongId.md
ranokay: Done.

content/en/docs/Endpoints/getLyricsBySongId.md
kgarner7: My bad, I appear to not be clear. I'm trying to understand _why_ this should be avoided in the first place (I am ok with `must` over `should`, as `should` is a weak word).

content/en/docs/Responses/cueLine.md
ranokay: Both points make sense — I reworked the design:

1. `role` becomes a clean enum (`bg`, `voice`, `group`, omitted for main) with a separate `voiceIndex` (0-based integer, only present when role is `voice`). This eliminates the `voiceN` string pattern.
2. `displayRole` added as an optional free-text string for servers that have named agent metadata (e.g. `ttm:name` in TTML). Omitted when source doesn't provide it.

kgarner7: Thanks! Would it be possible to include these fields in an example in the JSON?

content/en/docs/Endpoints/getLyricsBySongId.md
ranokay: Overlapping cues break word-by-word karaoke highlighting. If cue A runs from 1000–2000ms and cue B starts at 1500ms, the client has two cues "active" simultaneously — which one gets the highlight? Clients would need overlap-resolution logic (z-ordering, priority rules, etc.) instead of simply iterating cues sequentially and highlighting the current one.

For *source* data this is unavoidable — some formats have naturally overlapping timing. But when the server is *deriving* end-times that didn't exist in the source (e.g. filling in missing `end` values from Enhanced LRC), it's generating those boundaries from scratch, so it can trivially set `cue[n].end = cue[n+1].start` and produce clean, non-overlapping intervals. No reason to create a problem that doesn't exist in the source.

kgarner7: Thanks. So, within a given `cueLine->cue`, there should never be overlapping items? (but, for different roles, that would be allowed?)
If so, I think that makes sense to _require_ this from the server (then, there is nothing to do from the client).

content/en/docs/Responses/cueLine.md
ranokay: Added! The new example shows a multi-voice scenario with `role: "voice"` + `voiceIndex` for two individual singers and `role: "group"` for a chorus line. `displayRole` carries the singer names from TTML `ttm:name`.

content/en/docs/Endpoints/getLyricsBySongId.md
ranokay: Good point. Within a single `cueLine`, cues should never overlap. Cross-role overlaps (different cueLines with different `role`/`voiceIndex`) are expected since those represent parallel vocal layers. I'll promote this to a blanket **must**: servers must ensure cues within a cueLine don't overlap, regardless of whether end-times come from the source or are derived.

kgarner7 (dismissed):

content/en/docs/Responses/cueLine.md
Tolriq: Not sure we want to remove main if we want to name it with a display name. Main is voice 0 no ?

ranokay: Just to make sure I'm reading you right — are you saying that if the main singer is e.g. Chris Martin, they should use `role: "voice"`, `voiceIndex: 0` rather than omitted role, so we don't lose the ability to name them?

If so: `displayRole` doesn't require `role` to be set — it's an independent optional field. For a single-voice song where the source provides the singer name, the cueLine would be:

```json
{ "index": 0, "displayRole": "萬妮達", "cue": [...] }
```

No `role` needed — main vocals stay as the default (omitted). Making main = voice 0 would force every single-voice cueLine to carry `role: "voice", voiceIndex: 0`, which adds noise for the most common case.

In multi-voice scenarios (duets, ensembles), all voices use `role: "voice"` with `voiceIndex` — each part belongs to a specific singer.

ranokay: My intent was: empty/omitted `role` still denotes the default/main layer, even if `displayRole` is present. So a named lead can be represented as `{ "index": 0, "displayRole": "Chris Martin", ... }` and remain the main/default cueLine.

`voiceIndex` is only used for cueLines whose `role` is `"voice"`; it enumerates additional explicit voice layers, not the default/main layer.

If you’d prefer all named singers to be modeled uniformly, including the lead as `role: "voice", voiceIndex: 0`, I can change the docs/examples that way, but that would mean changing the current “omitted role = main” rule.`

Tolriq: But we do not want to duplicate the display role in all the cues right ?. That would be highly inefficient.

ranokay: Correct — I do **not** mean duplicating `displayRole` onto individual `cue` entries.

My intent is that `displayRole` is **cueLine-level metadata**, just like `role`. All cues inside that `cueLine` implicitly belong to that named vocal layer, so the name is carried once per `cueLine`, not repeated per cue.

So for example, a named lead/main layer would still be:

```json
{ "index": 0, "displayRole": "Chris Martin", "cue": [...] }
```
and an additional named voice layer would be another `cueLine` with its own `displayRole`.

Tolriq: Yes so it's still duplicated per line, I thought it would be an index mapping as ttml for efficiency.

If we really want those names, we should not repeat the same data and use the same efficient system as ttml. An agentId in the cueLine and an agents array at the top with id, name, role.

ranokay: That makes sense. I was only trying to avoid duplication on each `cue`, but you're right that `displayRole` still duplicates the same name on every `cueLine` across the track.

If we want named singer/character attribution in v2, an indirection model is cleaner: `agentId` on `cueLine` plus a shared `agents[]` collection, rather than repeating the same label per line.

I can rework it that way instead of `displayRole`. Before I change it, can you clarify two points so I match your intent correctly:

1. Should `agentId` replace both `voiceIndex` and `displayRole`?
2. Should `role` live only on the agent entry (`id`, `name`, `role`), or remain on `cueLine` as the normalized vocal-layer field?

Tolriq: agentId replaces all, the role, the voiceIndex and the displayRole. The displayRole should probably be agentName, the role stays and the voiceIndex is no more needed as the agentId share the same purpose.

ranokay: That makes sense to me. I read your intent as:

- `cueLine` carries `agentId`
- shared `agents[]` carries the reusable metadata
- `role`, `voiceIndex`, and `displayRole` are removed from `cueLine`

Before I rework the docs/schemas again, can you confirm these points so I match the model correctly:

1. Where should `agents[]` live: on each `structuredLyrics` entry, on `lyricsList`, or at the top response level?
2. For the default/main layer, is omitted `agentId` still the compact/default case, or should a named lead/main also always reference an agent?
3. What should the agent object be exactly: `{ id, name, role }` or `{ id, agentName, role }`? And for a named lead/main agent, should `role` stay omitted/default, stay `voice`, or do you want an explicit `main` role value?

Tolriq: I should probably run claude directly ;) I'm not a machine I just put ideas out loud to reach the best end.

Anyway question 1 makes no sense, it can only be an entry in each structuredLyrics.
The name can be name no need for agentName.

For the rest IMO 2 choices, either we always enforce explicit, and each line must have an agent, or we only enforce that as soon as there's 2 agent (including just a main and a chorus).

Role in the agent data should never be empty, we can either define a main or say first voice is main.

ranokay: That makes sense. I’ll rework the proposal around a shared `agents[]` collection on each `structuredLyrics` entry plus `agentId` on `cueLine`, and drop `role` / `voiceIndex` / `displayRole` from `cueLine`.

I’ve been using AI to help me iterate on the spec shape, and I’m also learning these modeling tradeoffs while doing it, so I appreciate the back-and-forth and the corrections.

Tolriq: You can use AI to build the spec and the code, but it's usually still better to actually write yourself when you talk to other people to not sound like a robot ;) Or change your llms tone to something more human ;)

ranokay: I pushed the rework.

I changed the model to use a shared `agents[]` list on each structuredLyrics entry plus `cueLine.agentId`, and removed role, voiceIndex, and displayRole from cueLine.

I also tightened the rules a bit:

- if `agents` are used for cue-attributed lyrics, there has to be exactly one `main` agent
- `agents[].id` has to be unique within that structuredLyrics entry
- every `agentId `has to point to one local agent in the same entry
- simple single-layer lyrics can still omit agents, but multi-agent / multi-layer entries now need them

Tolriq: Can you clarify `Agent identity is also independent across `kind` tracks. Clients should not assume an `agentId` in one `structuredLyrics` entry refers to the same singer, character, or vocal layer in another entry.`  ? since there's only 1 agents array and ids are unique, I don't see how that's possible.

ranokay: I meant `agents` are scoped per `structuredLyrics` entry, not globally across the whole response, so two separate entries could each have their own local `agents[]`.
But the per-entry scoping rule already covers that, so I’ll remove this extra sentence to avoid confusion.

kgarner7 (approved):

Tolriq:
I've mostly already validated as helped shape it. Waiting for at least one @opensubsonic/servers to confirm they don't see this as problematic to implement as it's the first evolution of a version of an extension.

epoupon (approved):

Tolriq (approved):

ayla6:
I found a small problem with the spec. It's very specific and wouldn't be found in any decent quality lyrics but:
If you have something like this

```xml
<span begin="00:00:00.000" end="00:00:00.500">Hello</span> <span begin="00:00:00.500" end="00:00:01.000">yay</span> hello <span begin="00:00:02.500" end="00:00:03.000">hello</span> <span begin="00:00:03.000" end="00:00:03.500">hi</span> <span begin="00:00:03.500" end="00:00:04.500">awesome</span>
```

where one word without synced data is followed by an identical one with synced data, it becomes impossible to tell which one of them is synced when it gets turned into the json the spec uses:

```json
"cueLine": [
            {
              "index": 0,
              "start": 0,
              "end": 5000,
              "value": "Hello yay hello hello hi awesome",
              "cue": [
                {
                  "start": 0,
                  "end": 500,
                  "value": "Hello"
                },
                {
                  "start": 500,
                  "end": 1000,
                  "value": "yay"
                },
                {
                  "start": 2500,
                  "end": 3000,
                  "value": "hello"
                },
                {
                  "start": 3000,
                  "end": 3500,
                  "value": "hi"
                },
                {
                  "start": 3500,
                  "end": 4500,
                  "value": "awesome"
                }
              ]
            }
          ],
```

it's minor but it's just something i've noticed

ayla6:
if i can give my two cents   maybe it should be changed to use start character and end character or something, it's simpler to apply and is a lot more accurate. probably should've said it before it being merged tho

Tolriq:
@ranokay seems your PR is still not merged in Navidrome so we can still amend to start end char positions. WDYT ?

ranokay:
Yeah, fair point, that edge case is genuinely ambiguous with the current model. I hadn’t considered the repeated untimed/timed identical-token case.
I’m okay with adjusting it before Navidrome merges. I do wonder if explicit ordered tokens/segments would age better than char offsets, but if we want the smallest possible amendment, positions on the cues are probably the simpler path.
I see two possible fixes:
1. Minimal amendment: keep `cue` as-is and add char positions into `cueLine.value`:
```json
{ "value": "hello", "start": 2500, "end": 3000, "charStart": 16, "charEnd": 21 }
```
2. Bigger but cleaner change: preserve explicit ordered tokens, timed and untimed:
```json
"tokens": [
  { "value": "Hello", "start": 0, "end": 500 },
  { "value": "yay", "start": 500, "end": 1000 },
  { "value": "hello" },
  { "value": "hello", "start": 2500, "end": 3000 }
]
```
My feeling is char positions are the smaller amendment, while ordered tokens are the cleaner long-term model. What do you think?

Tolriq:
We said earlier in the spec that when there's an end all cues (not token :p) must have them, so the 2 would contradict that and make things more complicated client side. Internally I do use solution 1 and I guess most will need to do the char mapping, so directly doing it is better for clients IMO.

ranokay:
Alright, I think the main thing to pin down is the exact semantics: relative to `cueLine.value`, 0-based or not, and whether charEnd is exclusive. We should probably also define what “character position” means for Unicode so different implementations don’t count differently.

If you’re aligned on that direction, I can put together a small follow-up PR for the schema/docs.

Tolriq:
0 based and inclusive the First "Hello" is 0-4, for Unicode that's a great question :) Considering the data to generate those endpoints can't have cue split in the middle of a unicode, using byte offset is probably the simplest to avoid issues for some clients ?

ranokay:
My only concern is naming: if we go with UTF-8 byte offsets, I think they should be called `byteStart` / `byteEnd` rather than `charStart` / `charEnd`, otherwise people will assume actual character positions.

Then we can define them precisely as 0-based inclusive offsets into the UTF-8 encoding of the final `cueLine.value` string, with no normalization step.

Tolriq:
works for me.

ranokay:
Opened #228

PR #219, opened 2026-03-12
adding similarTracks and findPath new api for AudioMuse-AI sonic simiarity

NeptuneHub:
This pr is to add `similarTracks` and `findPath` new  API for support **AudioMuse-AI Sonic Similarity** as discussed here https://github.com/opensubsonic/open-subsonic-api/discussions/172

Please let me know if I need to add or fix something.

Thanks.


NeptuneHub:
I renamed them `findSonicPath` and `getSonicSimilarTracks`.

Then to keep them deadly simple the Input parameter are the id of the song and the number of song in response.

Output parameter are id of the song and, just for getSonicSimilarTracks, the distance.

I really think that this is the bare minimum then to create update the queue and avoid API tailored on AudioMuse-AI. Hope that finally we can reach a consensus on this.

epoupon:
> The distance was in the result, so clients can better identify the data it get and can use it.
> 
> For example Symfonium have extra features for the mixes like excluded from mix value to for songs, or exclude recently skipped, this means that I always request more data than needed to be able to filter out those. But if I ask 20 similar to metal and it returns classical near the end, I can detect via the distribution of the distance, and remove those too far away.
> 
> So distance normalized to 0.0 1.0 is important for the client to take actions.

Yes, but normalized to what?
0.0 for the first song, 1.0 for the returned one which is the farthest?
If it is normalized regarding the whole library, I am afraid values will change from one library to another. If there are only metal songs in the library, you would eliminate legit values

NeptuneHub:
is something else missing ? Can we proceed with approving this PR to be merged ?

Tolriq:
You really need to chill down :) Always too pushy everywhere :)

You forget to answer to @epoupon and I think he's right we should make it even more agnostic.

NeptuneHub:
I think I already wrote some message above that because we made this API specific for Sonic Similarity, cosine similarity is the gold standard. So that I like to keep as it is.

content/en/docs/Endpoints/findsonicpath.md
```diff
@@ -0,0 +1,110 @@
...
+          "channelCount": 2,
+          "track": 4,
+          "year": 2007,
+          "genre": "Electronic",
+          "size": 792375,
+          "discNumber": 1,
+          "suffix": "wma",
+          "contentType": "audio/x-ms-wma",
+          "path": "Synthetic/Synthetic_-_Colorsmoke_EP/04-Synthetic_-_BrownSmoke.wma",
+          "distance": 0.0
```
epoupon: "similarity"

content/en/docs/Endpoints/findsonicpath.md
```diff
@@ -0,0 +1,110 @@
...
+          "channelCount": 2,
+          "track": 5,
+          "year": 2007,
+          "genre": "Electronic",
+          "size": 3209886,
+          "discNumber": 1,
+          "suffix": "mp3",
+          "contentType": "audio/mpeg",
+          "path": "Synthetic/Synthetic_-_Colorsmoke_EP/05-Synthetic_-_RedGreenSmoke.mp3",
+          "distance": 0.45
```
epoupon: "similarity"

content/en/docs/Endpoints/getsonicsimilartracks.md
```diff
@@ -0,0 +1,109 @@
...
+}
+{{< /tab >}}
+{{< tab header="Subsonic" lang="json" >}}
+Does not exist.
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `sonicSimilarTracksResult` | [`sonicSimilarTracksResult`](../../responses/sonicsimilartracksresult) | **Yes** |  | The sonic similar tracks result |
```
epoupon: why not directly the array of results btw?

content/en/docs/Responses/sonicMatch.md
```diff
@@ -0,0 +1,51 @@
...
+}
+{{< /tab >}}
+{{< tab header="Subsonic" lang="json" >}}
+Does not exist.
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| All [`Child`](../child) fields | | | | All fields from the [`Child`](../child) type. |
```
epoupon: For "bookmark", we have a child element field (entry) and additional fields at the same level (no inheritence). We should be consistent in the API

content/en/docs/Responses/sonicMatch.md
```diff
@@ -0,0 +1,51 @@
...
+| `similarity` | `number` | **Yes** |  | The normalized similarity score (1.0 = most similar, 0.0 = least similar). For `getSonicSimilarTracks`, relative to the query song. For `findSonicPath`, relative to the starting song. Returns -1 when similarity is not supported by the server. |
```
epoupon: What do you mean by "not supported by the server"? the server is able to give a path but it is not able to quantify the similarity score?

That being said, audio similarity should be an OS extension. Only servers that report support for that extension will implement these endpoints.

content/en/docs/Endpoints/findsonicpath.md
```diff
@@ -0,0 +1,110 @@
...
+}
+{{< /tab >}}
+{{< tab header="Subsonic" lang="json" >}}
+Does not exist.
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `findSonicPathResult` | [`findSonicPathResult`](../../responses/findsonicpathresult) | **Yes** |  | The sonic path result |
```
epoupon: why not directly the array of results btw?

content/en/docs/Endpoints/findsonicpath.md
NeptuneHub: solved

content/en/docs/Endpoints/findsonicpath.md
NeptuneHub: solved

content/en/docs/Responses/sonicMatch.md
NeptuneHub: This is a suggestion form @Tolriq that in case the Music Server (Navidrome or whatelse) use some kind of plugin that don't have the distance value it returned as -1.

content/en/docs/Endpoints/findsonicpath.md
NeptuneHub: Right, I removed the extra findSonicPathResult wrapper

content/en/docs/Endpoints/getsonicsimilartracks.md
NeptuneHub: Right, I removed the extra sonicsimilartracksresult wrapper

content/en/docs/Responses/sonicMatch.md
NeptuneHub: So instead of this:
```
{
  "sonicMatch": [
    {
      "id": "300000060",
      "parent": "200000002",
      "title": "BrownSmoke",
      "isDir": false,
      "isVideo": false,
      "type": "music",
      "albumId": "200000002",
      "album": "Colorsmoke EP",
      "artistId": "100000002",
      "artist": "Synthetic",
      "coverArt": "300000060",
      "duration": 304,
      "bitRate": 20,
      "bitDepth": 16,
      "samplingRate": 44100,
      "channelCount": 2,
      "track": 4,
      "year": 2007,
      "genre": "Electronic",
      "size": 792375,
      "discNumber": 1,
      "suffix": "wma",
      "contentType": "audio/x-ms-wma",
      "path": "Synthetic/Synthetic_-_Colorsmoke_EP/04-Synthetic_-_BrownSmoke.wma",
      "similarity": 1.0
    },
..
  ]
}
```

you suggest this to be aligned to bookmark:
```
{
  {
    "sonicMatch": [
      "entry": {
        "id": "300000060",
        "parent": "200000002",
        "title": "BrownSmoke",
        "isDir": false,
        "isVideo": false,
        "type": "music",
        "albumId": "200000002",
        "album": "Colorsmoke EP",
        "artistId": "100000002",
        "artist": "Synthetic",
        "coverArt": "300000060",
        "duration": 304,
        "bitRate": 20,
        "bitDepth": 16,
        "samplingRate": 44100,
        "channelCount": 2,
        "track": 4,
        "year": 2007,
        "genre": "Electronic",
        "size": 792375,
        "discNumber": 1,
        "suffix": "wma",
        "contentType": "audio/x-ms-wma",
        "path": "Synthetic/Synthetic_-_Colorsmoke_EP/04-Synthetic_-_BrownSmoke.wma",
      }, 
      "similarity": 1.0
  },
..
  ]
}
```

I personally prefear my implementation in this case. Are there any other reason to do that more than "already made in bookmark"?

Tolriq: Consistency ? Reusability of Child without duplication for auto generated clients ? API surface are not really built around personal preferences ;)

NeptuneHub: Ok changed as suggested.

Tolriq:
As @epoupon said do not forget to add the extension part (https://opensubsonic.netlify.app/docs/extensions/template/)

NeptuneHub:
> As @epoupon said do not forget to add the extension part (https://opensubsonic.netlify.app/docs/extensions/template/)

Added.

epoupon (dismissed):

content/en/docs/Responses/sonicSimilarTracksResult.md
```diff
@@ -0,0 +1,79 @@
...
+  The result of a sonic similar tracks search.
+---
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "sonicMatch": [
+    {
+      "id": "300000060",
```
Tolriq: The sample is wrong it does not show the new sonicMatch.

content/en/docs/Responses/findSonicPathResult.md
```diff
@@ -0,0 +1,79 @@
...
+  The result of a sonic path search between two songs.
+---
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "sonicMatch": [
+    {
+      "id": "300000060",
```
Tolriq: The sample is wrong it does not show the new sonicMatch.

content/en/docs/Endpoints/findsonicpath.md
```diff
@@ -0,0 +1,113 @@
...
+}
+{{< /tab >}}
+{{< tab header="Subsonic" lang="json" >}}
+Does not exist.
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `sonicMatch` | Array of [`sonicMatch`](../../responses/sonicmatch) | **Yes** |  | The ordered sonic path results with per-track similarity |
```
Tolriq: We need to define more precisely if the path contains the start and end songs or not. (Plex does for example so we need to be explicit about that.)

content/en/docs/Responses/sonicSimilarTracksResult.md
NeptuneHub: Thanks for spotting this, fixed.

content/en/docs/Responses/findSonicPathResult.md
NeptuneHub: Thanks for spotting this, fixed.

content/en/docs/Endpoints/findsonicpath.md
NeptuneHub: Fixed: path must contain the start song in the beginning and the end song in the end.

Tolriq (dismissed):

epoupon (dismissed):

epoupon:
@opensubsonic/servers  @opensubsonic/clients  we need another vote please :)

content/en/docs/Endpoints/getsonicsimilartracks.md
```diff
@@ -0,0 +1,112 @@
...
+- Addition
+categories:
+- Browsing
+description: >
+  Returns similar tracks based on sonic (audio) analysis.
+---
+
+`http://your-server/rest/getSonicSimilarTracks`
+
+Returns tracks that are sonically similar to a given track, based on audio analysis. Each result includes its normalized distance from the query track.
```
deluan: I know this was discussed in the PR and associated discussion, but I think the spec needs to be crystal clear what this endpoint returns. We should not mention `distance` here. Also, what similarity == 0 means? In the example below it suggests that lower is more similar, the examples have similarity 0.05 and 0.12, suggesting the first one is more similar but have a lower value. Is that what was agreed upon?

deluan (changes requested):
Looks good. Just a small issue

content/en/docs/Endpoints/getsonicsimilartracks.md
NeptuneHub: Hi Deluan,
thanks for spotting this: I replaced `distance` with `similarity` also here. I also search in the entire documentation and there is no any other `distance` leftover.

You also spotted another things that I correct:
- when this value was called `distance` we have 0 meaning the same identical song (like same song twice in the collection), 1 the most different one.
- now we called it `similarity` and is totally the contrary (like similarity = 1 - distance). If you have 1 mean you have the same identical song.

I fixed both.

NeptuneHub:
Addressed also the point from Deluan. So now when we talk about similarity

* 1 is the max, means that is the same song duplicated in the library;
* 0 is the min, means totally different song;

I also did a search to avoid other `distance `wording missed.

Please re-review :)

NeptuneHub:
Also aligned the example of the two API to show that first responses are the most similar to similarity 1.0

deluan (dismissed):

Tolriq (dismissed):

content/en/docs/Responses/findSonicPathResult.md
```diff
@@ -0,0 +1,83 @@
...
+}
+{{< /tab >}}
+{{< tab header="Subsonic" lang="json" >}}
+Does not exist.
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `sonicMatch` | Array of [`sonicMatch`](../sonicmatch) | **Yes** |  | An ordered list of songs forming the path, each with its similarity score from the starting song. |
```
paulijar: IMO this page `findSonicPathResult.md` as well as the page `sonicSimilarTracksResult.md` have no reason to exist in the documentation. The corresponding endpoint pages don't even link to these response types, they link directly to the `sonicMatch.md` (which is fine).

These two `...Result.md` files are also basically identical with each other as they are just arrays of `sonicMatch` entries. So an alternative would be to combine them into one page called `sonicMatches.md` but I would find even that to be a distraction compared to linking directly to `sonicMatch.md`.

content/en/docs/opensubsonic-versions.md
```diff
@@ -12,7 +12,7 @@ This table shows the supported extensions by OpenSubsonic versions:
 
 | OpenSubsonic version | Supported extension |
 | --- | --- |
-| 1 | [Template extension](../extensions/template) |
+| 1 | [Template extension](../extensions/template), [Sonic similarity](../extensions/sonicSimilarity) |
```
paulijar, line 15: I'm not certain what is the point of this table, but none of the other extensions are listed here. So it doesn't seem quite right to have the Sonic similarity here alone.

Maybe we should ditch this entire page as long as there's no concrete plans of an API v2? But that's of course an issue to be discussed elsewhere.

paulijar (changes requested):

epoupon (dismissed):
Thanks!

content/en/docs/Responses/findSonicPathResult.md
NeptuneHub: Thanks for spotting this, I just removed `findSonicPathResult.md` and `sonicSimilarTracksResult.md`.

They are a lefthover from one of the requested change.

content/en/docs/opensubsonic-versions.md
NeptuneHub: I added this because was asked from two reviewer:

_As @epoupon said do not forget to add the extension part (https://opensubsonic.netlify.app/docs/extensions/template/)_

NeptuneHub:
Just addressed the comment from @paulijar removing 2 result file unused, leftover from previous change.

Is meant that each change need a new review from everyone? if yes, please re-review. thanks :)

content/en/docs/opensubsonic-versions.md
paulijar: Yes, adding the extension page was important, but linking it to this `opensubsonic-versions.md` is something which hasn't been done for the other 8 extensions which we have. But it's no big deal, really.

paulijar (approved):

Tolriq (approved):

epoupon (approved):

epoupon:
Hello!
I’ve made quite a bit of progress on integrating a custom similarity engine into LMS based on acoustic similarities. So far, so good.
However, I’m noticing that the results from getSonicSimilarTracks tend to heavily favor tracks from the same album as the source track... and this is somewhat expected.
I think this could be a problem for the client, which ends up having to discard quite a lot of results before it can suggest something that’s meaningfully different from the original track.
For LMS, I’ve built a playlist generator on top of the recommendation engine: it skips/rearranges the similarity results to enforce constraints similar to what you’d find on “real” radio (no two tracks from the same album back-to-back, no more than N tracks from the same album within a sliding one-hour window, same idea for artists, etc.).
So I’m not entirely sure what the best approach is:
* expose new endpoints for this, more “playlist-oriented” rather than raw acoustic results. But note there is also the same problem if we want sonic similar albums, we may want to filter out albums from the same artist.
* getSonicSimilarTracks endpoints take into account for this server-side, though that may feel a bit clunky since you don’t always want filtering
* or leave it up to clients to handle... But then each client will likely reinvent this logic, with varying quality...

What do you think?

NeptuneHub:
What I did on AudioMuse-AI is DISCARD more than X song, where X is a configurable constant, from the same Author. By default 3.

So this means that the client will never get more than 3 songs from the same author in a result, resolving also the problem of multiple song of the same album.

I also avoid duplicate automatically, because maybe you have the same song duplicated in multiple album.

Tolriq:
This will always depend on how the algo works, for Plex this is actually not the case similarity does not really favor album or artists on a large enough database.

epoupon:
Ok thanks for your answers!
So basically option 2: you rely on the server to expose only filtered stuff. Will do then!

Th3K1ngP1n:
To everyone involved with this. I come here as a mere user, and, just wanted to really thank you all for the hard work and dedication for making this possible. Just last year, I remember wishing so much for this to be possible, and, here we are. The combination of Symfonium, Navidrome, and, Audiomuse is truly magnificent. (To all other OpenSub music servers, I mean no harm to your work, this is just the stack I use)

PR #228, opened 2026-04-02
feat(songLyrics): add cue byte offsets

ranokay:
## Summary

This follow-up amends the `songLyrics` v2 cue model to disambiguate repeated timed/untimed text within a `cueLine.value`.

It adds optional `byteStart` / `byteEnd` fields on `cue`, defined as 0-based inclusive UTF-8 byte offsets into the final parent `cueLine.value`, with no normalization step.

## Changes

- add `byteStart` / `byteEnd` to the `cue` docs and OpenAPI schema
- document them as an all-or-none pair
- document that they are only valid relative to `cueLine.value`
- document the invariant `byteStart <= byteEnd`
- propagate the contract text through `cueLine`, `structuredLyrics`, `songLyrics`, and `getLyricsBySongId`

### DIFF
--- a/content/en/docs/Endpoints/getLyricsBySongId.md
+++ b/content/en/docs/Endpoints/getLyricsBySongId.md
@@ -23,7 +23,7 @@ The lyrics can come from embedded tags (`SYLT`/`USLT`), LRC file/text file, or a
 | Parameter  | Req.    | OpenS.  | Default | Comment                                                                                                                                                |
 | ---------- | ------- | ------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `id`       | **Yes** | **Yes** |         | The track ID.                                                                                                                                          |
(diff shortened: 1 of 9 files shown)


Tolriq:
Do we want that optional ?

ranokay:
Maybe keep them optional in the schema, but conditionally required by the spec?

If cue order + `cue.value` is enough to map cues back to `cueLine.value` unambiguously, offsets can be omitted. If not, then `byteStart` / `byteEnd` must be present for the relevant cues.

That fixes the repeated timed/untimed token case without forcing every server to emit byte offsets for all cues.

Tolriq:
The thing is that if the server needs to be able to handle this case it means it can emit those. Making that optional means that servers must find if there's duplicates or not and if emitting or not, and clients needs to handle both the case with or without those.

This only brings complexity on both ends for no gain, it's actually better to enforce it so every one handle only one case.

ranokay:
Gotcha. If we do that, `cueLine.value` should probably also become required whenever `cueLine.cue` is present, since the byte offsets are defined relative to that exact string. Otherwise the offsets are mandatory but their reference text is still optional.

Tolriq:
It's already required in your spec no ?

ranokay:
No, not currently. It’s only conditionally required when byte offsets are emitted.

Tolriq:
I looked at the wrong value. Yes value should be required

Tolriq (approved):

ranokay:
@epoupon @kgarner7 could you take a look?

Tolriq (approved):

kgarner7 (approved):

paulijar (approved):

PR #230, opened 2026-04-09
feat(extension): infiniteLibrary — scalable support for large and cloud-backed catalogs

soyelmismo:
## Summary

This PR introduces the `infiniteLibrary` OpenSubsonic extension to address scalability and determinism challenges that arise when a server is backed by a very large or cloud-hosted music catalog (e.g., Tidal, Qobuz, or self-hosted collections with 100K+ items).

## Motivation

Several existing endpoints were designed with the assumption that the server has a small, fully-indexed local library. When libraries grow to millions of items, or when the server proxies an external catalog:

- **String-based artist lookups become ambiguous** — `getTopSongs(artist="Savant")` could match multiple unrelated artists. Meanwhile, `getSimilarSongs2` correctly uses an `id` parameter. This inconsistency is addressed.
- **Returning all result types wastes resources** — A client searching only for albums must still trigger artist and song searches. The behavior of `artistCount=0` is undefined across implementations.
- **N+1 cover art fetching** — Displaying 25 search results requires 25 additional `getCoverArt` calls when direct URLs could be embedded.
- **Binary-only streaming** — Cloud services deliver audio via HLS/DASH manifests, forcing proxy servers to stitch segments. This is CPU-intensive and adds latency.
- **Empty query sync is impractical** — Returning "all data" for infinite catalogs is not feasible.

### New extension file
- `Extensions/infiniteLibrary.md` — Formal extension definition (v1/v2/v3)

### Modified endpoints
- **`getTopSongs`**: Added optional `artistId` parameter for deterministic ID-based resolution
- **`search3`**: Added `type` filter parameter, clarified zero-count semantics
- **`stream`**: Added `streamMode` parameter (`binary`/`redirect`/`manifest`)

### Modified responses
- **`searchResult3`**: Added `totalArtistCount`, `totalAlbumCount`, `totalSongCount` fields for proper pagination
- **`AlbumID3`**: Added `coverArtUrl` field for direct CDN artwork URLs
- **`error`**: Added error code `71` (result set too large)

## Design Principles

1. **All changes are optional** — Servers declare support via `getOpenSubsonicExtensions`
2. **All new parameters have defaults matching current behavior** — No existing client breaks
3. **All new response fields are additive** — Clients that don't understand them ignore them
4. **Follows the existing extension template format exactly**

## Real-World Validation

These proposals come from practical implementation experience building [Substream](https://github.com/soyelmismo/substream), a Subsonic-compatible proxy for Tidal with 100M+ items. Each issue was discovered through real client behavior (Feishin, Aonsoku, Supersonic) connecting to cloud-backed catalogs.

## Complexity Analysis

| Operation | Current Protocol | With Extension |
|---|---|---|
| Artist top songs lookup | O(n) fuzzy search | O(1) ID lookup |
| Search (artist-only) | 3 parallel searches | 1 targeted search |
| Cover art (25 results) | 25 HTTP calls | 0 calls (URL embedded) |
| Stream initiation | 2-5s (proxy stitch) | <100ms (redirect) |

### DIFF
--- a/content/en/docs/Endpoints/gettopsongs.md
+++ b/content/en/docs/Endpoints/gettopsongs.md
@@ -15,8 +15,17 @@ Returns top songs for the given artist, using data from [last.fm](http://last.fm
 
 | Parameter | Req. | OpenS. | Default | Comment |
 | --- | --- | --- | --- | --- |
-| `artist` | **Yes** |  |   | The artist name. |
+| `artist` | **Yes*** |  |   | The artist name. |
+| `artistId` | No | **Yes** |   | The artist ID. If provided, takes precedence over the `artist` string. |
 | `count` | No  | |  50  | Max number of songs to return. |
+
+{{< alert color="warning" title="OpenSubsonic" >}}
+**`infiniteLibrary` extension (version 1):**
+
+The `artist` parameter requires the server to perform a string-based lookup which can be ambiguous when multiple artists share similar names. Servers that support the [`infiniteLibrary`](../../extensions/infinitelibrary/) extension SHOULD accept the `artistId` parameter and use it for a direct, deterministic lookup.
+
+When `artistId` is provided, the `artist` parameter becomes optional. Servers MUST return an error if neither `artist` nor `artistId` is provided.
+{{< /alert >}}
 
 ### Example
 
--- a/content/en/docs/Endpoints/search3.md
+++ b/content/en/docs/Endpoints/search3.md
@@ -27,9 +27,16 @@ Music is organized according to ID3 tags.
 | `songCount` | No |   | 20  | Maximum number of songs to return. |
 | `songOffset` | No |   | 0   | Search result offset for songs. Used for paging. |
 | `musicFolderId` | No |   |     | (Since [1.12.0](../../subsonic-versions)) Only return results from music folder with the given ID. See `getMusicFolders`. |
+| `type` | No | **Yes** |  | Comma-separated list of result types to include: `artist`, `album`, `song`. If omitted, all types are returned. |
 
 {{< alert color="warning" title="OpenSubsonic" >}}
 Servers must support an **empty query** and return all the data to allow clients to properly access all the media information for offline sync.
+
+**`infiniteLibrary` extension (version 1):**
+
+- Servers that support very large catalogs MAY return error code `71` for empty queries when the library exceeds a practical sync threshold.
+- The `type` parameter allows clients to explicitly request only the result types they need. When specified, the server SHOULD skip computation for excluded types entirely (not just omit them from the response). This is semantically stronger than setting a count to `0`, which has undefined behavior across implementations.
+- When a count parameter is explicitly set to `0` (e.g., `artistCount=0`), servers MUST NOT return results of that type, and SHOULD NOT perform the underlying search for that type.
 {{< /alert >}}
 
 ### Example
--- a/content/en/docs/Endpoints/stream.md
+++ b/content/en/docs/Endpoints/stream.md
@@ -26,6 +26,7 @@ Streams a given media file.
 | `size` | No  |   |   | (Since [1.6.0](../../subsonic-versions)) Only applicable to video streaming. Requested video size specified as WxH, for instance "640x480". |
 | `estimateContentLength` | No |  | false | (Since [1.8.0](../../subsonic-versions)). If set to "true", the *Content-Length* HTTP header will be set to an estimated value for transcoded or downsampled media. |
 | `converted` | No  | | false | (Since [1.14.0](../../subsonic-versions)) Only applicable to video streaming. Servers can optimize videos for streaming by converting them to MP4. If a conversion exists for the video in question, then setting this parameter to "true" will cause the converted video to be returned instead of the original. |
+| `streamMode` | No | **Yes** | `binary` | `binary`: current behavior (raw audio bytes). `redirect`: server returns HTTP 302 to the audio source URL. `manifest`: server returns an HLS/DASH manifest if the source supports adaptive streaming. `infiniteLibrary` extension (version 2). |
 
 ### Example
 
@@ -40,4 +41,12 @@ OpenSubsonic servers **must not** count access to this endpoint as a play and in
 
 If the server support the [`Transcode Offset`](../../extensions/transcodeoffset/) extension, then it must accept the `timeOffset` parameter for music too.
 
+**`infiniteLibrary` extension (version 2):**
+
+The `streamMode` parameter allows clients to opt into more efficient streaming modes:
+
+- `redirect`: The server responds with HTTP 302 pointing to the actual audio file URL. Suitable for native clients that can follow redirects and handle direct streams. This eliminates server-side proxying overhead entirely.
+- `manifest`: The server responds with an HLS (`.m3u8`) or DASH (`.mpd`) manifest. Suitable for clients that support adaptive bitrate streaming. The response Content-Type will be `application/vnd.apple.mpegurl` or `application/dash+xml`.
+
+If the server cannot fulfill the requested `streamMode`, it MUST fall back to `binary` mode silently.
 {{< /alert >}}
--- a/content/en/docs/Extensions/infiniteLibrary.md
+++ b/content/en/docs/Extensions/infiniteLibrary.md
@@ -1,0 +1,126 @@
+---
+title: "Infinite Library"
+linkTitle: "Infinite Library"
+OpenSubsonic:
+- Extension
+description: >
+  Enables scalable, deterministic behavior for servers backed by very large or cloud-hosted music catalogs.
+---
+
+**OpenSubsonic version**: [1](../../opensubsonic-versions)
+
+**OpenSubsonic extension name**: `infiniteLibrary` (As returned by [`getOpenSubsonicExtensions`](../../endpoints/getopensubsonicextensions))
+
+This extension addresses scalability and determinism challenges that arise when an OpenSubsonic server acts as a bridge to a very large music catalog (e.g., cloud-hosted services with millions of items, or large self-hosted collections with ambiguous metadata).
+
+The core principle is: **prefer IDs over strings, and allow clients to request only what they need.**
+
+## Motivation
+
+Several existing endpoints were designed with the assumption that the server has a small, fully-indexed local library. When libraries grow to millions of items, or when the server proxies an external catalog:
+
+- String-based artist lookups become ambiguous (e.g., multiple artists named "Savant")
+- Returning all three result types (artists, albums, songs) for every search wastes resources when the client only needs one
+- Clients cannot distinguish exact matches from fuzzy ones
+- Cover art proxying creates unnecessary load when direct URLs are available
+
+## Version 1
+
+### `getTopSongs` — ID-based artist resolution
+
+The `getTopSongs` endpoint currently requires an `artist` string parameter. This extension adds an optional `artistId` parameter that, when provided, takes precedence over the `artist` string and eliminates ambiguity.
+
+**New parameter:**
+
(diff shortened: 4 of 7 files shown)


Tolriq:
Random LLM rambling here with nearly no coherence and logic, specially the 3 version of the extension ...

dweymouth:
@soyelmismo The biggest problem IME as a client developer with the current API for very large libraries is the inability to paginate on endpoints like `getArtists`, `getStarred2`, `getPlaylists` and `getPlaylist` (plus the lack of a `getSongList` API but that is somewhat tangential). Proposing an extension to fix that using similar paging to `getAlbumList2` would probably get a lot more traction

Tolriq:
We certainly need the new endpoints with filtering/sorting/paging that I might tacle one day, but others can do all the work too and contribute more on all the prs and discussions ;)

soyelmismo:
> would probably get a lot more traction

Yeah, but basically the entire extension is needed for the future, unless we don't want OpenSubsonic to standardize big catalogs at all.

Pagination is just one part of the structural issue, and it's an easy fix. But we need query granularity just as much as we need pagination :)

I can split the PR and start with a pagination extension first if that gets things moving.

soyelmismo:
> Random LLM rambling, specially the 3 version of the extension ...

The v3 extension isn't rambling at all, it's based on real server architectures.

OpenSubsonic handles multiple users, right? Tracking what's played in common across the entire server perfectly justifies a native trending sort type.

And the libraryType flag (local/remote/hybrid) is literally for servers acting as a bridge to platforms like Tidal. Think of how HiBy Music Player integrates Tidal, but done 2-in-1 directly through the OpenSubsonic protocol. If clients don't know the server is remote or hybrid, they try to force full-library offline syncs or scans that simply crash.

Tolriq:
The thing is that you did not took 10 seconds to see how the extension versions works, there's no reason to add 3 version of a new extension on the first pass ... You create an extension that's all, if the 3 things are different things then they are different extensions ....

And yes currently open subsonic is about your music that you have indexed already. You need stable ids and many things, that requires consistency.

Paging is also not so easy to add as there's no point in breaking the current endpoints. Paging without sorting is also kinda useless on large dataset and since we need new endpoints we also need to add support for field selection to optimize client and server communication.

There's already a lot of discussions and things and thinking around OS it's goal and trajectory. Not just a fork of a well known server with 10 quick claude PR trying to rework a complex public API with all the design impact this have.

soyelmismo:
We’ve deciphered his message. His ego is bruised because he noticed the structured “AI” format (*Claude PR*) and is on the defensive, but he also gave clear technical reasons:

1. **Our rookie mistake:** He’s right that you don’t submit extensions labeled “Versions 1, 2, and 3” on day one. It should have been a single base proposal or three separate extensions. 
2. **His vision:** He literally says that *Open Subsonic is for music that YOU already have indexed* with stable IDs on your hard drive. He doesn’t want dynamic proxies within the official standard.
3. **Pagination:** He believes that pagination also requires fixing the sorting and sending completely new endpoints.

**The Verdict:**
He is protecting the *status quo* of the original project. Debating here is a waste of time on bureaucracy.

**Tactical Action Plan (Selfish Path):**
Close the PR yourself or leave it ignored. Don’t respond any further.
Build the ecosystem and philosophy in **your own fork and in Tempus**. If the Tempus community moves to Substream because it simply works wonderfully with Tidal, the conventional OpenSubsonic protocol will become obsolete on its own.

Talk is cheap; code rules. Substream already works. Let’s stop fighting with them and keep optimizing our backend

Translated with [DeepL.com](https://www.deepl.com/?utm_campaign=product&utm_source=web_translator&utm_medium=web&utm_content=copy_free_translation) (free version)


>- that was funny, but yeah, since we can't still bring our cloud library into our own local databe, good luck with the opensubsonic protocol :)

dweymouth:
> Paging is also not so easy to add as there's no point in breaking the current endpoints. Paging without sorting is also kinda useless on large dataset and since we need new endpoints we also need to add support for field selection to optimize client and server communication.

Adding `offset` and `count` as new params on the existing endpoints isn't breaking and is very straightforward with an extension. No reason not to get this one out as a quick win IMO, just to bring the endpoints up to the standard that `getAlbumList2` sets

soyelmismo:
> Adding `offset` and `count` as new params on the existing endpoints isn't breaking 

Indeed. Actually, the whole extension is 100% non-breaking and additive. Old clients and servers simply ignore the new fields they don't map. It's just a quick win for huge datasets without messing with the classic local folder setups.

Tolriq:
**Pagging without a proper API is very inefficient with large dataset,** Navidrome have hack after 50k items in search3 else the performance are horrible due to how most database works. That's why for very large dataset we usually have the concept of continuation tokens for example. So yes we can add an extension that will introduce something that have not a lot of purpose (you can already use the search3 endpoint with paging for the most important parts) and will not be efficient.

There's many different end points, adding paging in a single extension, would mean a server needs to support all the endpoints with paging at once (including maybe endpoints it does not yet support) or we need to define a ton of extensions all that for something very very limited.

Edit: Again this was discussed a couple of times.

soyelmismo:
> Edit: Again this was discussed a couple of times.

So, if an efficient offset/pagination solution is demonstrated for local libraries, would you be open to standardize it?

Strictly speaking, OpenSubsonic defines the API schema contract, not the internal backend algorithms. If the API contract is solid, server devs can figure out their own indexing solutions (sqlite, nosql, etc). However, since you have deep experience on the client side, what specific paging structure would you suggest for endpoints like getArtists to avoid the pitfalls you mentioned?

dweymouth:
> you can already use the search3 endpoint with paging for the most important parts

Actually as a replacement for `getArtists`, probably. I should try that. But not for `getPlaylist` to fetch playlists that may be very large with 10,000s or more tracks (esp. smart playlists that may pull in the whole library or a large subset of it into one playlist). Not for `getStarred2` either since search3 has no way to specify fetching only starred items

Tolriq:
> So, if an efficient offset/pagination solution is demonstrated for local libraries, would you be open to standardize it?

Again this is a more global problem, specially around the full number of items, and generic enough solutions without breaking older clients, hence the new endpoints, typically your new coverUrl that does not even match the album details one that can have multiple size. This means a lot of extra data for the server to gather and network, hence the need to control the fields we request to optimize everything.

> Actually as a replacement for getArtists, probably. I should try that. But not for getPlaylist to fetch playlists that may be very large with 10,000s or more tracks (esp. smart playlists that may pull in the whole library or a large subset of it into one playlist)

While getPlaylist is one of the rare where an extension would make sense as no client side sorting to request, it's also typically one of the case that absolutely require a continuation token, a smart playlist sorted by random via paging without proper handling is exactly the kind of things we can't properly handle on a stateless API.

dweymouth:
> Strictly speaking, OpenSubsonic defines the API schema contract, not the internal backend algorithms. If the API contract is solid, server devs can figure out their own indexing solutions (sqlite, nosql, etc). However, since you have deep experience on the client side, what specific paging structure would you suggest for endpoints like getArtists to avoid the pitfalls you mentioned?

New endpoints with pagination by a continuation token (instead of offset) and the ability to specify sort/filter options in the call would be the eventual goal. Tolriq and I are in alignment there. Where there is some disagreement is whether there is some value in getting an easier stop-gap solution out first that will be replaced with the new endpoints, or just wait until we have a plan for redesigning all the browsing endpoints.

soyelmismo:
Exactly. A stop-gap solution is all I'm advocating for.

What if we restrict this stop-gap strictly to deterministic endpoints (like getArtists) where offset works flawlessly? We drop the complex stuff like getPlaylist so we don't introduce random-sorting bugs.

It fixes the 50k+ crashing issue for clients today while you draft the perfect token-based endpoints for the future. I can open a highly scoped PR just for that.


PS: Honestly, the main trigger for this PR was that I currently can't browse massive artist discographies—or search specific albums in cloud backends—via Supersonic without hitting that hard limit of returned albums. A stop-gap pagination solves that exact pain point today 😁

Tolriq:
You can already via search3 that allows pagination of artists and albums ... search3 with empty value is deterministic and fit your need.

soyelmismo:
`search3` does not accept an `artistId` filter in the current schema; it relies on string matching.

If a client wants to securely load a specific artist's massive discography by their unique ID, it has to call `getArtist`. But `getArtist` dumps everything at once with no pagination bounds.

If we try to hack search3 by sending the artist's name just to get paginated albums, we get text-matching cascades and false positives from other artists with similar names.

Secure, ID-based pagination without string-matching hacks is exactly what this stop-gap extension provides for deterministic endpoints like getArtist.

Tolriq:
I missed the no s version. Anyway IMO this is also exactly the kind of endpoints that requires sorting. From my little user experience, users sorts the artist albums by release date or original release or year or album type.

Having pagination just to be able to load large dataset because the server can't serialize the data is usually not the proper solution here.

dweymouth:
Is `getArtist` really an endpoint that we expect to need pagination though? The most prolific artists out there have maybe, on the order of a few hundred albums in their discography? That should be fine in one go and client-side sorting post-API call. If we eventually introduce an endpoint to get all songs for an artist directly (or it is done through a new `getSongs` endpoint with a filter option by artist ID(s)) then we would need pagination.

dweymouth:
> PS: Honestly, the main trigger for this PR was that I currently can't browse massive artist discographies—or search specific albums in cloud backends—via Supersonic without hitting that hard limit of returned albums. A stop-gap pagination solves that exact pain point today 😁

If Supersonic crashes (not times out, but crashes) with loading large discographies, please open an issue, that shouldn't be happening. Tiemout is adjustable in the config file as well, and I'd accept a PR to make it configurable in UI via the Advanced tab of the settings dialog

soyelmismo:
> If Supersonic crashes (not times out, but crashes) 

@dweymouth You hit the nail on the head. Supersonic didn't natively crash; it timed out.

Because the protocol demands full album metadata hydration (durations, track counts, etc.), translating a cloud API artist response (like Tidal's) into a Subsonic response required an N+1 fetching loop per album on my backend.

Fetching 300+ albums caused massive network timeouts. So I had to artificially hard-cap the proxy server to max 10-50 albums per query just to avoid API rate limits and looong several minutes of loading from my end. That's why the pagination was a matter of life or death...

dweymouth:
> > If Supersonic crashes (not times out, but crashes)
> 
> @dweymouth You hit the nail on the head. Supersonic didn't natively crash; it timed out.
> 
> Because the protocol demands full album metadata hydration (durations, track counts, etc.), translating a cloud API artist response (like Tidal's) into a Subsonic response required an N+1 fetching loop per album on my backend.
> 
> Fetching 300+ albums caused massive network timeouts. So I had to artificially hard-cap the proxy server to max 10-50 albums per query just to avoid API rate limits and looong several minutes of loading from my end. That's why the pagination was a matter of life or death...

In this case, the stop-gap pagination without the ability to specify server-side sorts won't solve anything, since Supersonic sorts (client side) discographies by album name or release year. To switch to a paginated endpoint it'd need the sort options as well, so the full replacement endpoint would be needed.

Tolriq:
In french we usually then finish by CQFD ;)

soyelmismo:
CQFD indeed 😂.

If Supersonic is strictly dependent on dumping the entire catalog for client-side sorting, then a simple offset patch here is completely useless anyway.

I'll forget about standard protocols for a day. I might just go bother @dweymouth over on the Supersonic repo directly about those cascading getArtist hydration loops instead 👋.

Thanks for the heavy debate!

paulijar:
Regarding the part of the OP about the HLS streaming, there already is the endpoint https://opensubsonic.netlify.app/docs/endpoints/hls/ in the original Subsonic API definition. But I'm not certain if there actually are any servers or clients out there implementing this endpoint.

PR #232, opened 2026-04-14
Add groupings field

Morikko:
Add an optional `groupings` field to the `Child` response.

It is the representation of:

|  Unified | ID3v2.3 | ID3v2.4 |  MP4 | ASF |  Vorbis  |
|:--------:|:-------:|:-------:|:----:|:---:|:--------:|
| Grouping | GRP1    | GRP1    | ©grp |     | GROUPING |

https://kid3.sourceforge.io/kid3_en.html#frame-list

The grouping tag is used to categorize tracks into a broader group beyond just the album. I have not found an accurate description but people on the web started to use it for various use cases.

It is common for me to use multiple values for the field (through Navidrome string splitter) so I decided to go with an array. I agree with the same motivation from https://github.com/opensubsonic/open-subsonic-api/pull/212#issuecomment-3933781662: It is still possible to just use an array of size 1 to simulate a single value.

A few examples of my usage:
- The music source: CD, Bandcamp, Qobuz
- The music type: Christmas, Soundtrack, Kid (i do not find genres would fit)

An discussion is also already opened: https://github.com/opensubsonic/open-subsonic-api/discussions/182. Although without an answer, it received a few upvotes already.

### DIFF
--- a/content/en/docs/Responses/Child.md
+++ b/content/en/docs/Responses/Child.md
@@ -135,7 +135,8 @@ description: >
       "number": 2,
       "count": 4
     }
-  ]
+  ],
+  "groupings": ["Soundtrack", "Live"]
 }
 {{< /tab >}}
 {{< tab header="Subsonic" lang="json" >}}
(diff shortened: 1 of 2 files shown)


kgarner7 (approved):

kgarner7:
@opensubsonic/servers @opensubsonic/clients This should be pretty safe to add, would you please help review?

deluan (approved):

epoupon (approved):

PR #237, opened 2026-05-09
Add collections

ygabuev:
Follows the discussion in https://github.com/opensubsonic/open-subsonic-api/discussions/73.

Add the concept of a collection with basic CRUD functionality. A collection is an ordered list of items - songs, artist, albums, playlists (other types can be added in the future). Users can create, modify and delete collections.

The proposed endpoints and schemas closely follow those for playlists, however, items don't have to be just songs. Additionally, the `updateCollection` endpoint has a more elaborate mechanism for adding, removing and moving items around via the `add`, `remove`, and `move` parameters respectively.

New methods:
- `getCollections`
- `getCollection`
- `createCollection`
- `updateCollection`
- `deleteCollection`

The existing methods and responses all stay the same.

There are no clear contribution guidelines, so sorry if I opened the PR prematurely without clarifying the details in the discussion. I figured that it's easier to discuss the proposal against a more concrete OpenAPI implementation.

Open questions:
- [ ] do we need more item types (podcast, internet radio, ...)?
- [ ] should users be able to share collections?
- [ ] cover art for collections?

### DIFF
--- a/content/en/docs/Endpoints/createcollection.md
+++ b/content/en/docs/Endpoints/createcollection.md
@@ -42,6 +42,10 @@ The request payload should be provided in the body as a JSON object.
     {
       "type": "album",
       "id": "200000021"
+    },
+    {
+      "type": "genre",
+      "name": "vaporwave"
     }
   ]
 }
@@ -68,7 +72,7 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a top-le
       "comment" "this is a collection",
       "owner": "user",
       "public": false,
-      "itemCount": 2,
+      "itemCount": 3,
       "created": "2023-03-16T03:18:41+00:00",
       "changed": "2023-03-16T03:18:41+00:00",
       "items": [
@@ -119,6 +123,13 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a top-le
             "artist": "Comfort Fit",
             "year": 2005,
             "genre": "Hip-Hop"
+          }
+        },
+        {
+          "genre": {
+            "value": "vaporwave",
+            "songCount": 6,
+            "albumCount": 1
           }
         }
       ]
--- a/content/en/docs/Endpoints/createcollection.md
+++ b/content/en/docs/Endpoints/createcollection.md
@@ -16,21 +16,10 @@ This endpoint must be accessed using an HTTP POST request.
 
 ### Request Body
 
-The request payload should be provided in the body as a JSON object.
-
-| Parameter | Req. | OpenS. | Default | Comment |
-| --- | --- | --- | --- | --- |
-| `name` | **Yes** |  |  | The human-readable name of the collection. |
-| `comment` | No |  |  | The collection comment. |
-| `public` | No |  | `false`  | `true` if the collection should be visible to all users, `false` otherwise. |
-| `items` | No |  |  | A list of [collectionItemID](../payloads/collectionitemid) objects. |
-
-### Example request
-
-{{< alert color="primary" >}} `POST http://your-server/rest/createCollection.view?u=demo&p=demo&v=1.16.1&c=AwesomeClientName&f=json` {{< /alert >}}
+The request payload should be provided in the body as a JSON object, as specified by the [CreateCollectionRequest](../payloads/createcollectionrequest) schema.
 
 {{< tabpane persist=false >}}
-{{< tab header="Request body" lang="json">}}
+{{< tab header="CreateCollectionRequest" lang="json">}}
 {
   "name": "test collection",
   "comment": "this is a collection",
@@ -51,6 +40,10 @@ The request payload should be provided in the body as a JSON object.
 }
 {{< /tab >}}
 {{< /tabpane >}}
+
+### Example request
+
+{{< alert color="primary" >}} `POST http://your-server/rest/createCollection.view?u=demo&p=demo&v=1.16.1&c=AwesomeClientName&f=json` {{< /alert >}}
 
 ### Result
 
--- a/content/en/docs/Endpoints/createcollection.md
+++ b/content/en/docs/Endpoints/createcollection.md
@@ -1,0 +1,133 @@
+---
+title: "createCollection"
+linkTitle: "createCollection [OS]"
+categories:
+- Collections
+opensubsonic:
+- Extension
+description: >
+  Creates a collection.
+---
+
+`http://your-server/rest/createCollection`
+
+Creates a collection.
+This endpoint must be accessed using an HTTP POST request.
+
+### Request Body
+
+The request payload should be provided in the body as a JSON object.
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `name` | **Yes** |  |  | The human-readable name of the collection. |
+| `comment` | No |  |  | The collection comment. |
+| `public` | No |  | `false`  | `true` if the collection should be visible to all users, `false` otherwise. |
+| `items` | No |  |  | A list of [collectionItemID](../payloads/collectionitemid) objects. |
+
+### Example request
+
+{{< alert color="primary" >}} `POST http://your-server/rest/createCollection.view?u=demo&p=demo&v=1.16.1&c=AwesomeClientName&f=json` {{< /alert >}}
+
+{{< tabpane persist=false >}}
+{{< tab header="Request body" lang="json">}}
+{
+  "name": "test collection",
+  "comment": "this is a collection",
+  "items": [
+    {
+      "type": "song",
+      "id": "300000060"
+    },
+    {
+      "type": "album",
+      "id": "200000021"
+    }
+  ]
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+### Result
+
+A [`subsonic-response`](../../responses/subsonic-response) element with a top-level [`collection`](../../responses/collection) object on success.
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "subsonic-response": {
+    "status": "ok",
+    "version": "1.16.1",
+    "type": "AwesomeServerName",
+    "serverVersion": "0.1.3 (tag)",
+    "openSubsonic": true,
+    "collection": {
+      "id": "800000075",
+      "name": "test collection",
+      "comment" "this is a collection",
+      "owner": "user",
+      "public": false,
+      "itemCount": 2,
+      "created": "2023-03-16T03:18:41+00:00",
+      "changed": "2023-03-16T03:18:41+00:00",
+      "items": [
+        {
+          "song": {
+            "id": "300000060",
+            "parent": "200000002",
+            "title": "BrownSmoke",
+            "isDir": false,
+            "isVideo": false,
+            "type": "music",
+            "albumId": "200000002",
+            "album": "Colorsmoke EP",
+            "artistId": "100000002",
+            "artist": "Synthetic",
+            "coverArt": "300000060",
+            "duration": 304,
+            "bitRate": 20,
+            "bitDepth": 16,
+            "samplingRate": 44100,
+            "channelCount": 2,
+            "userRating": 5,
+            "averageRating": 5,
+            "track": 4,
+            "year": 2007,
+            "genre": "Electronic",
+            "size": 792375,
+            "discNumber": 1,
+            "suffix": "wma",
+            "contentType": "audio/x-ms-wma",
+            "path": "Synthetic/Synthetic_-_Colorsmoke_EP-20k217-2007/04-Synthetic_-_BrownSmokeYSBM20k22khS.wma"
+          }
+        },
+        {
+          "album": {
+            "id": "200000021",
+            "parent": "100000036",
+            "album": "Forget and Remember",
+            "title": "Forget and Remember",
+            "name": "Forget and Remember",
+            "isDir": true,
+            "coverArt": "al-200000021",
+            "songCount": 20,
+            "created": "2021-07-22T02:09:31+00:00",
+            "duration": 4248,
+            "playCount": 0,
+            "artistId": "100000036",
+            "artist": "Comfort Fit",
+            "year": 2005,
+            "genre": "Hip-Hop"
+          }
+        }
+      ]
+    }
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `collection` | [`CollectionWithItems`](../../responses/collectionwithitems) | **Yes** |   | The collection |
--- a/content/en/docs/Endpoints/createcollection.md
+++ b/content/en/docs/Endpoints/createcollection.md
@@ -48,6 +48,7 @@ The request payload should be provided in the body as a JSON object, as specifie
 ### Result
 
 A [`subsonic-response`](../../responses/subsonic-response) element with a top-level [`collection`](../../responses/collection) object on success.
+In case of an error, a standard HTTP error code is returned with a descriptive message.
 
 {{< tabpane persist=false >}}
 {{< tab header="**Example**:" disabled=true />}}
--- a/content/en/docs/Endpoints/createcollection.md
+++ b/content/en/docs/Endpoints/createcollection.md
@@ -68,65 +68,7 @@ In case of an error, a standard HTTP error code is returned with a descriptive m
       "public": false,
       "itemCount": 3,
       "created": "2023-03-16T03:18:41+00:00",
-      "changed": "2023-03-16T03:18:41+00:00",
-      "items": [
-        {
-          "song": {
-            "id": "300000060",
-            "parent": "200000002",
-            "title": "BrownSmoke",
-            "isDir": false,
-            "isVideo": false,
-            "type": "music",
-            "albumId": "200000002",
-            "album": "Colorsmoke EP",
-            "artistId": "100000002",
-            "artist": "Synthetic",
-            "coverArt": "300000060",
-            "duration": 304,
-            "bitRate": 20,
-            "bitDepth": 16,
-            "samplingRate": 44100,
-            "channelCount": 2,
-            "userRating": 5,
-            "averageRating": 5,
-            "track": 4,
-            "year": 2007,
-            "genre": "Electronic",
-            "size": 792375,
-            "discNumber": 1,
-            "suffix": "wma",
-            "contentType": "audio/x-ms-wma",
-            "path": "Synthetic/Synthetic_-_Colorsmoke_EP-20k217-2007/04-Synthetic_-_BrownSmokeYSBM20k22khS.wma"
-          }
-        },
-        {
-          "album": {
-            "id": "200000021",
-            "parent": "100000036",
-            "album": "Forget and Remember",
-            "title": "Forget and Remember",
-            "name": "Forget and Remember",
-            "isDir": true,
-            "coverArt": "al-200000021",
-            "songCount": 20,
-            "created": "2021-07-22T02:09:31+00:00",
-            "duration": 4248,
-            "playCount": 0,
-            "artistId": "100000036",
-            "artist": "Comfort Fit",
-            "year": 2005,
-            "genre": "Hip-Hop"
-          }
-        },
-        {
-          "genre": {
-            "value": "vaporwave",
-            "songCount": 6,
-            "albumCount": 1
-          }
-        }
-      ]
+      "changed": "2023-03-16T03:18:41+00:00"
     }
   }
 }
@@ -135,4 +77,4 @@ In case of an error, a standard HTTP error code is returned with a descriptive m
 
 | Field |  Type | Req. | OpenS. | Details |
 | --- | --- | --- | --- | --- |
-| `collection` | [`CollectionWithItems`](../../responses/collectionwithitems) | **Yes** |   | The collection |
+| `collection` | [`Collection`](../../responses/collection) | **Yes** |   | The collection |
--- a/content/en/docs/Endpoints/createcollection.md
+++ b/content/en/docs/Endpoints/createcollection.md
@@ -34,7 +34,7 @@ The request payload should be provided in the body as a JSON object, as specifie
     },
     {
       "type": "genre",
-      "name": "vaporwave"
+      "id": "vaporwave"
     }
   ]
 }
--- a/content/en/docs/Endpoints/getcollection.md
+++ b/content/en/docs/Endpoints/getcollection.md
@@ -18,6 +18,8 @@ Returns a collection with the items it contains.
 | Parameter | Req. | OpenS. | Default | Comment |
 | --- | --- | --- | --- | --- |
 | `id` | **Yes** |  |  | ID of the collection to return, as obtained by `getCollections`. |
+| `itemCount` | No |  |  | The number of items from the collection to return. If unset or set to a negative value, return all items. |
+| `itemOffset` | No |  |  | The number of items to skip. |
 
 ### Example
 
--- a/content/en/docs/Endpoints/getcollection.md
+++ b/content/en/docs/Endpoints/getcollection.md
@@ -1,0 +1,108 @@
+---
+title: "getCollection"
+linkTitle: "getCollection [OS]"
+categories:
+- Collections
+OpenSubsonic:
+- Extension
+description: >
+  Returns a collection.
+---
+
+`http://your-server/rest/getCollection`
+
+Returns a collection with the items it contains.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `id` | **Yes** |  |  | ID of the collection to return, as obtained by `getCollections`. |
+
+### Example
+
+{{< alert color="primary" >}} `http://your-server/rest/getCollection.view?id=123&u=demo&p=demo&v=1.13.0&c=AwesomeClientName&f=json` {{< /alert >}}
+
+### Result
+
+A [`subsonic-response`](../../responses/subsonic-response) element with a top-level `collection` object of type [collectionWithItems](../../responses/collectionwithitems)
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "subsonic-response": {
+    "status": "ok",
+    "version": "1.16.1",
+    "type": "AwesomeServerName",
+    "serverVersion": "0.1.3 (tag)",
+    "openSubsonic": true,
+    "collection": {
+      "id": "800000075",
+      "name": "testcreate",
+      "comment" "this is a collection",
+      "owner": "user",
+      "public": true,
+      "itemCount": 2,
+      "created": "2023-03-16T03:18:41+00:00",
+      "changed": "2023-03-16T03:18:41+00:00",
+      "items": [
+        {
+          "song": {
+            "id": "300000060",
+            "parent": "200000002",
+            "title": "BrownSmoke",
+            "isDir": false,
+            "isVideo": false,
+            "type": "music",
+            "albumId": "200000002",
+            "album": "Colorsmoke EP",
+            "artistId": "100000002",
+            "artist": "Synthetic",
+            "coverArt": "300000060",
+            "duration": 304,
+            "bitRate": 20,
+            "bitDepth": 16,
+            "samplingRate": 44100,
+            "channelCount": 2,
+            "userRating": 5,
+            "averageRating": 5,
+            "track": 4,
+            "year": 2007,
+            "genre": "Electronic",
+            "size": 792375,
+            "discNumber": 1,
+            "suffix": "wma",
+            "contentType": "audio/x-ms-wma",
+            "path": "Synthetic/Synthetic_-_Colorsmoke_EP-20k217-2007/04-Synthetic_-_BrownSmokeYSBM20k22khS.wma"
+          }
+        },
+        {
+          "album": {
+            "id": "200000021",
+            "parent": "100000036",
+            "album": "Forget and Remember",
+            "title": "Forget and Remember",
+            "name": "Forget and Remember",
+            "isDir": true,
+            "coverArt": "al-200000021",
+            "songCount": 20,
+            "created": "2021-07-22T02:09:31+00:00",
+            "duration": 4248,
+            "playCount": 0,
+            "artistId": "100000036",
+            "artist": "Comfort Fit",
+            "year": 2005,
+            "genre": "Hip-Hop"
+          }
+        }
+      ]
+    }
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `collection` | [`collectionWithItems`](../../responses/collectionwithitems) | **Yes** |   | The collection |
--- a/content/en/docs/Endpoints/getcollection.md
+++ b/content/en/docs/Endpoints/getcollection.md
@@ -43,7 +43,7 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a top-le
       "comment" "this is a collection",
       "owner": "user",
       "public": true,
-      "itemCount": 2,
+      "itemCount": 3,
       "created": "2023-03-16T03:18:41+00:00",
       "changed": "2023-03-16T03:18:41+00:00",
       "items": [
@@ -94,6 +94,13 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a top-le
             "artist": "Comfort Fit",
             "year": 2005,
             "genre": "Hip-Hop"
+          }
+        },
+        {
+          "genre": {
+            "value": "vaporwave",
+            "songCount": 6,
+            "albumCount": 1
           }
         }
       ]
--- a/content/en/docs/Endpoints/getcollections.md
+++ b/content/en/docs/Endpoints/getcollections.md
@@ -1,0 +1,71 @@
+---
+title: "getCollections"
+linkTitle: "getCollections [OS]"
+OpenSubsonic:
+- Extension
+categories:
+- Collections
+description: >
+  Returns all collections a user has access to.
+---
+
+`http://your-server/rest/getCollections`
+
+Returns all collections a user has access to.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `username` | No |  |  | If specified, return collections for this user rather than for the authenticated user. The authenticated user must have admin role if this parameter is used. |
+
+### Example
+
+{{< alert color="primary" >}} `http://your-server/rest/getCollections.view?u=demo&p=demo&v=1.13.0&c=AwesomeClientName&f=json` {{< /alert >}}
+
+### Result
+
+A [`subsonic-response`](../../responses/subsonic-response) element with a top-level `collections` array of [collection](../../responses/collection) elements on success.
+
+{{< tabpane persist=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+{
+  "subsonic-response": {
+    "status": "ok",
+    "version": "1.16.1",
+    "type": "AwesomeServerName",
+    "serverVersion": "0.1.3 (tag)",
+    "openSubsonic": true,
+    "collections": [
+      {
+        "id": "800000075",
+        "name": "testcollection",
+        "owner": "user",
+        "public": true,
+        "created": "2026-03-16T03:18:41+00:00",
+        "changed": "2026-03-16T03:18:41+00:00",
+        "itemCount": 3,
+        "readonly": true,
+        "validUntil": "2026-03-23T03:18:41+00:00"
+      },
+      {
+        "id": "800000076",
+        "name": "testcollection2",
+        "comment" "this is another collection",
+        "owner": "user",
+        "public": false,
+        "itemCount": 17,
+        "created": "2026-03-16T03:18:41+00:00",
+        "changed": "2026-03-16T03:18:41+00:00",
+        "readonly": true
+      }
+    ]
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `collections` | Array of [`collection`](../../responses/collection) | **Yes** |   | The collections |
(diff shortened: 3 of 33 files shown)


Tolriq:
It must be an extension, markdown first before openAPI.

With that said your API does not work for the ordered part, arrays of content per type does not allow mixed content ordering, the solution also does not properly handle duplicates and moving stuff.

Should be an array of items with mediaType mediaId probably.

ygabuev:
@Tolriq thanks for the comment. I will provide the markdown docs soon.

You are right, the per-type arrays are wrong, I was confused by the existing semantics for parameters in the GET versions of `createX` endpoints. I've added a dedicated `CollectionItemId` schema with `type` and `id` fields (the word "media" is a bit out of place here IMO, e.g. artists are not media). Those should be used for creating and updating collections.

I've also added the extension comments and errors for the new endpoints

ygabuev:
Hey @Tolriq. I am reworking the proposal and would appreciate some input.

I see that many of the older endpoints that manipulate data have a very particular way of using HTTP methods, e.g. `createPlaylist` allows `GET`, and `deleteUser` allows `GET` and `POST` (but not `DELETE`). Also, many body payloads require `application/x-www-form-urlencoded` MIME type. This quickly becomes unwieldy for complex payloads, particularly when arrays of objects are supplied. For example, creating a collection with a payload similar to
```json
{
  "items": [
    {"type": "song", "id": "1"},
    {"type": "album", "id": "5"}
  ]
}
```
would require a pretty weird encoding that is - as far as I have checked - not well specified, or at least very convoluted.

I wonder if it makes sense to
- use `POST` for `createCollection`, `PUT` for `updateCollection`, `DELETE` for `deleteCollection`; not use `GET`
- use `application/json` in creation and update requests?

I see that you went with this approach in [a recent change](https://github.com/opensubsonic/open-subsonic-api/pull/168), so at least there is a precedent of not following the legacy Subsonic approaches.

Tolriq:
New endpoints should use json it was added for that and move away from the limitation of original API.

ygabuev:
I've modified the proposal a bit and also added the markdown pages, now it's `application/json` all the way. Would appreciate another look and more comments.

lachlan-00:
collection should support everything which i think would be

* album
* artist
* ~~bookmark~~ bookarks are just entries so skip them
* genre
* internetradiostation
* playlist
* podcastepisode
* podcast
* share
* song
* user
* video

if you're going to enforce PUT and DELETE why add separate endpoints? it would only make sense if you were going to allow GET on all endpoints as well.

everything could just be on a collection endpoint if we are moving extensions to this style

Tolriq:
There's no PUT and DELETE in the rest of the API there's no reason to introduce that now.

For collections, I do not see why it should contain bookmark or users or share.

lachlan-00:
okay then take it out of the delete requirement for the endpoint https://github.com/opensubsonic/open-subsonic-api/pull/237/changes#diff-431be3340a9769c0f1c11bf62185826558cf6b1eb4de14bb6103e0f0db672539R14

bookmarks are just entries so they can be skipped but shares and users are unique enough to stay for me.

they're not as useful but it would be good to support everything possible

Tolriq:
Shares can expire so does not make sense. Users are on most servers not visible from other users so also does not make sense.

lachlan-00:
okay then don't add them

ygabuev:
I've added the following types:
- genre
- internet radio station
- podcast episode
- podcast

Didn't add videos because they are not well defined in the API, would rather leave this as future work.

As a necessity, updated the `collectionItemId` type to support either `id` or `name` field because not all types support IDs, e.g. genre.

Also updated the HTTP method usage to only use GET and POST, in line with the rest of the API.

aweathe:
My biggest issue is the idea that a collection can be *anything*, people keep saying it but it just doesn't make sense to me. As an ex-API developer, I always followed the guiding principal of separation of concerns. Songs, albums, genres and artists all have intrinsically different data. Combining any API call would cause chaos from a frontend perspective and likely hurt the legibility of the schema - everything now needs to handle IF/ELSE/ELSEIF/SWITCH scenarios. The best scenario is almost always separation.

As an end goal I'd like album collections to be a first class citizen. eg.

Albums - All albums
Collections - Collated lists of albums
Favourites - Favourite Tracks/Albums
Playlists - Mix-tapes of favourite tracks

- A collection of songs is just a playlist. Artists and genres would be far better and easier implemented as tags. 
- A collection of albums is the only example that has a clear purpose outside of the default usages and a clear separation of concerns.
- A mix of songs and albums would be a headache for any frontend developer, which is why I err on the side of separation.

Tags aren't really the answer for album collections as they will always just end up in some generic tags section of UI (worked on many APIs and this usually happens) - album collections need to be a first class citizen like Albums, Artists, Favourites and Playlists.

Think about the frontend developer and how much of a pain that if/then/else/switch will be if it combines data types.

Tolriq:
What about thinking about user's need ? ;) A collection is a group of things. If you collect pokemons your collection can have cards, lego, books, .... And it still completely make sens ;)

aweathe:
> What about thinking about user's need ? ;) A collection is a group of things. If you collect pokemons your collection can have cards, lego, books, .... And it still completely make sens ;)

If you want a feature to have uptake by a wider number of clients and devices, the ease of implementation will dictate that. An endpoint that returns 5 or 6 types is asking to be pushed to the bottom of a list. A feature with a single datatype is easier to implement and creates less spagetti in the code.

Your pokemon is a great example of how the API should be laid out - an API should separate concerns, just like code and a contract should return a single, actionable datatype. Trying to pickup "my whole collection" is asking for trouble haha.

What it SHOULD return while still making everyone happy is:

/collections/cards <CardType>
/collections/books<BookType>
/collections/lego<LegoType>
/collection/sstuffed<StuffedType>

/collections should at MOST return a list of available types, e.g card, book, lego, stuffed so the app can know what endpoints it can call for what features. Then developers can do what they want and implement each feature as they see fit.

A single endpoint returning multiple types is asking for trouble

Tolriq:
Again not really, specially if you want ordering of the collection data ...

We are not talking about adding multiple tags on items and filtering on that tag, we are talking about ordering a collection.

For the Pokemon example, I want my shelf to show my collection ordered by year, so it will mix cards and lego and books in a specific order.

We define an API to fit a need, not the opposite way.

aweathe:
> Again not really, specially if you want ordering of the collection data ...
> 
> We are not talking about adding multiple tags on items and filtering on that tag, we are talking about ordering a collection.
> 
> For the Pokemon example, I want my shelf to show my collection ordered by year, so it will mix cards and lego and books in a specific order.
> 
> We define an API to fit a need, not the opposite way.

But in what way is a collection of songs, genres, albums, radio stations and artists filling a need? What's the use-case for that? I'm genuinely curious and the pokemon example doesn't work anymore, we're talking about music. What's the need? How does it play out? What does it look like to a user?

aweathe:
Being clear, I want this feature, I just want it to make sense and actually be added to clients so I can use it (yours as awell @Tolriq ).

Tolriq:
This is actually a frequent request on Symfonium to have collections of pinned different ordered items in rows.

epoupon:
> > Again not really, specially if you want ordering of the collection data ...
> > We are not talking about adding multiple tags on items and filtering on that tag, we are talking about ordering a collection.
> > For the Pokemon example, I want my shelf to show my collection ordered by year, so it will mix cards and lego and books in a specific order.
> > We define an API to fit a need, not the opposite way.
> 
> But in what way is a collection of songs, genres, albums, radio stations and artists filling a need? What's the use-case for that? I'm genuinely curious and the pokemon example doesn't work anymore, we're talking about music. What's the need? How does it play out? What does it look like to a user?

Same feeling here, I would be very curious about the actual use cases for mixing all this heterogeneous stuff

aweathe:
The GetStarred endpoint already does this with multiple object types, so I'm bowing out lol. There's no point in arguing API semantics when there's already an endpoint that does it. I hope to see it in the clients soon and not just buried in a "tags" or "collections" screen. My use case stands that I want to see collections of albums, as in Billboard Top 50 1982 followed by a list of album covers. As long as it does that, I'm fine with it.

https://opensubsonic.netlify.app/docs/endpoints/getstarred/

Tolriq:
You can downvote facts, this won't change that many actual users request that ... It's the pinned items of Spotify and other apps but not limited to one pinned list.

And whatever solution is chosen, most clients will probably have a collection screen that list collection and you can enter, what other way do you expect ? Else it's actually tags applied and smart filters but you loose ordering.

aweathe:
> You can downvote facts, this won't change that many actual users request that ... It's the pinned items of Spotify and other apps but not limited to one pinned list.
> 
> And whatever solution is chosen, most clients will probably have a collection screen that list collection and you can enter, what other way do you expect ? Else it's actually tags applied and smart filters but you loose ordering.

I didn't downvote facts, I downvoted that it's a "frequent request" - if there were supported facts in your statement, and a legitimate recurring use-case I would have given it a thumbs up. I asked a question and received a "because" instead.

aweathe:
Spotify is a great example, I can follow an artist, favourite an album (or add it to a folder), I can favourite a song.. where is this screen where I can do everything?

Tolriq:
> I didn't downvote facts, I downvoted that it's a "frequent request" - if there were supported facts in your statement, and a legitimate recurring use-case I would have given it a thumbs up. I asked a question and received a "because" instead.

So you have access to my mails and searched the forum for the opened feature requests to know that I invent things and that it's not already prepared inside the app because I obviously love to add things that no user wants ;) ? 

image

image

As I said facts .... 

And for spotify as said it's the pin stuff https://community.spotify.com/t5/Live-Ideas/Increase-the-amount-of-pins-available-in-Your-Library/idc-p/6667151 so you can see that many people use and want more of that ;)

aweathe:
> And for spotify as said it's the pin stuff https://community.spotify.com/t5/Live-Ideas/Increase-the-amount-of-pins-available-in-Your-Library/idc-p/6667151 so you can see that many people use and want more of that ;)

I'm not saying you're making stuff up, I'm saying that I asked a question and it was downplayed. I've read 6 pages of the spotify comments so far and everyone is asking to be able to pin more playlists rather than the standard 4. On the symfonium feature requests I see people asking for collections of albums or smart filters similar to plex but nothing like what we're talking about here lol. I'm just not seeing facts that people want a place for _everything_ collections.

BUT like I said, it's your API, do what you want - it already doesn't follow any meaningful standard regarding single responsibility with several of the endpoints, so there's no point in making this one do it either.

Tolriq:
This is not my API at all :) It's an attempt to fix an existing API and improve it while keeping backward compat and having all servers working exactly the same not like before.

We work with what we have ;)

And downplayed how ? You asked if there's a need I tell you that many users requested it ... So much that it's actually already prepared database side since a long time and I just need to find some time for the UI part.

For Spotify the pin is not only playlist that's the whole point, you can pin artists, playlists, tracks, and people want more than 4 not just for playlists. I've just posted on of the many posts showing that the pin stuff of mixed content is something used.

So 1) the API already work like that, 2) there's actually real users needs here even if you don't believe it. So there's no reason to cripple an API to remove things that users want to something different.

You still have not answered on what UI you imagine for collections that would not be a collection list while still allowing ordering so that the TOP 50 albums are in the proper order.

aweathe:
> You still have not answered on what UI you imagine for collections that would not be a collection list while still allowing ordering so that the TOP 50 albums are in the proper order.

UI for this is already challenging. 

**A collection of albums**: I can imagine people want to order them and the best way would be grpahically based on album covers, you could also view that as a table and move them that way, I guess. So you're at a multiple views for a single data type. For my use-case I think a sort would be more than adequate, some people may want more fine-grained control, I mean we can't do it with any other data type right now, so why this? I haven't looked at yours, but none of my other players allow me to order artists, albums, genres or even tracks using any kind of UI - just sorts.

**A collection of playlists**: Pretty similar to albums, you show playlist covers (or their names as a dummy image) similar to spotify. Easy as.

**A collection of songs**: Well, this is called a playlist

**A collection of genres**: Meh, could be useful, maybe, but also easily done at a tag level, or as a smart playlist. I'd more want the output of combined genres, not the list itself.

A collection of EVERYTHING.. there's no other way to do that than as a table and now all my album art has gone to waste. I mean you could join playlists and albums, but single tracks and genres wouldn't fit into a display-art paradigm. Like I said, this means that collections of albums and collections of playlists are once again second-class citizens.

This is all I'm arguing for, the collections endpoint makes perfect sense for all of these cases EXCEPT everything. You're building the app? How would this look to you?

Tracks - tabular data
Playlists - Tabular or image
Albums - Tabular or image
Genres - Tabular (usually just tag-list)
Artists - Tabular or artist image
Everything - Tabular? Again, that just loses the whole reason for having a well tagged library with great artwork.

You can only blend types if they have a same/similar view mode - images with images, otherwise it will just end up tabular which is crap for displaying artwork.

Tolriq:
I have no idea why you imagine that the display mode is impacted by the media type ? Everything can be displayed in every mode, specially in Symfonium case.

>  I mean we can't do it with any other data type right now, so why this?

Hum playlists ? That's the whole concept actually here. 

A collection is an ordered mixed media type content. What you want is a playlist of albums. There's no reason to limit that to albums when there's actual needs (Yes I know not actual real ones ;)) to have sorted mixed content.


Anyway the question is not about how to display the content of the collection that's actually not a problem at all, the question is how you want to list those collections and navigate them, because you wrote :

> I hope to see it in the clients soon and not just buried in a "tags" or "collections" screen.

And that's the part that makes no sense as I have no idea how you'd expose collections with sorted content in any other ways.
The playlists are in a playlist navigation screen ...

aweathe:
I'll let others chime in, I'm kinda out of steam with this and don't really have any reason to argue. You seem to think you have it sorted (and you probably do), however, I think it's likely based on a use-case that does not fit my own and we're arguing at cross-purposes.

Tolriq:
This is not about arguing ... This is about you exposing your needs so that clients can actually implement them ?

You can't say I want clients to do X but not like Y and when asked then how not answer. Then when clients do Y that sounds logical you come back and say that's not what I want. Because again I have absolutely no idea how any client currently could fit you need of album collections without showing a collection screen to navigate ...

This thing does fit your needs perfectly and support more things so no we are not cross purpose here.

aweathe:
> could fit you need of album collections without showing a collection screen to navigate ...

I mean not sourcing it as a _table view._ In most mixed-mode database crud/style applications, that usually ends up as a table. That doesn't show off my album covers.

Same as tags.. tags are rarely (in a UI perspective) shown as anything but tables or tag lists.

As I said that the beginning, my goal was this:

- Track
- Album
- ~AlbumCollection~
- Collections
  - Album
  - Playlist
  - Artist
- Artist
- Playlists
- Genres

Perhaps you see a collection like a WIKI page or something with multiple types? IF that's the case, then that *could* work. BUT you may as well return the types in separate parts of the response body.

```
Albums
[ ] [ ] [ ] [ ] [ ]

Playlists
[ ] [ ] [ ] [ ] [ ]

Tracks
- Track 1
- Track 2
``` 

Both of these are fine for a collections page (IMHO)
```
Table Layout ✔️ 
Collection 1
Collection 2
Collection 3

Image Layout ✔️ 
[C1] [C2] [C3]
```

What I cannot see is this which is what the API is returning (a page showing mixed types in a specified order):

- Album1
- Track 1
- Playlist 1
- Album2
- Track 2
- Track 3

Tolriq:
Tracks, playlists, artists, radio, ... have images right ?

So you can absolutely display a table of mixed content or a grid of mixed content, I have no idea why you find that impossible ?

Clients can add a small overlay for the type of each entry if mixed, or even different image handling like round for artists, or whatever they want. But the fact that it's mixed or not have 0 impact on what's possible client side.

Edit: Anyway you have clarified the important part that was not clear from your comments, so everything is already ready and planned at least in one client ;)

aweathe:
> Edit: Anyway you have clarified the important part that was not clear from your comments, so everything is already ready and planned at least in one client ;)

A client that I already paid for LOL

Tolriq:
with that said @ygabuev can you add a coverArt to the collection ?

ygabuev:
> with that said @ygabuev can you add a coverArt to the collection ?

I think users should be able to add their own cover art to collections, but I don't see any examples in API of how this should be done. If there were an approach in the API, it should also cover playlists, internet radio stations, podcasts, possibly avatars for users etc. How is this done for most clients and servers now?

Added the `covertArt` field for now, I assume this comes out-of-band.

Also added the `itemCount` and `itemOffset` parameters in `getCollection` for pagination.

Tolriq:
coverArt for now comes from the server, client side metadata edition is a larger more complex issue with security impacts via malformed images, too large images, ... That's another complex discussion for another time :)

content/en/docs/Endpoints/updatecollection.md
```diff
@@ -0,0 +1,92 @@
...
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `collectionId` | **Yes** |  |  | The collection ID. |
+| `name` | No |  |  | The human-readable name of the collection. |
+| `comment` | No |  |  | The collection comment. |
+| `public` | No |  |  | `true` if the collection should be visible to all users, `false` otherwise. |
+| `add` | No |  |  | Add the specified items to the collection. The payload is an array of [collectionItemID](../payloads/collectionitemid) objects. Items are appended to the tail of the collection. |
+| `move` | No |  |  | Move items from the original position to a new one in the collection. The original range is specified by the `fromStart` (inclusive) and `fromEnd` (exclusive) positions. The first item in the range moves to the position specified by `to` |
+| `remove` | No |  |  | Remove the items at the specified positions from the collection. The payload is an array of integer indices. |
```
kgarner7: All of these are missing type definitions. Also, `move` is missing a type definition in the payloads.

content/en/docs/Endpoints/createcollection.md
```diff
@@ -0,0 +1,144 @@
...
+  ]
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+### Result
+
+A [`subsonic-response`](../../responses/subsonic-response) element with a top-level [`collection`](../../responses/collection) object on success.
+
+{{< tabpane persist=false >}}
```
kgarner7, line 53: Would it make sense to give the _option_ to not return the collection body on create?

content/en/docs/Responses/collection.md
```diff
@@ -0,0 +1,48 @@
...
+| `id` | `string` | **Yes** |     | Id of the collection |
+| `name` | `string` | **Yes** |     | Name of the collection |
+| `comment` | `string` | No|     | A commnet |
+| `owner` | `string` | No |     | Owner of the collection |
+| `public` | `boolean` | No|     | Is the collection public |
+| `itemCount` | `int` | **Yes** |     | number of items |
+| `created` | `string` | **Yes** |     | Creation date [ISO 8601] |
+| `changed` | `string` | **Yes** |     | Last changed date [ISO 8601] |
+| `coverArt` | `string` | No |     | A cover Art Id |
+| `allowedUser` | Array of `string` | No |     | A list of allowed usernames |
```
kgarner7: What is the point of the `allowedUser` field? It's not specified as _configurable_ anywhere. Is the intention that it is _only_ server implementaiton specific?

content/en/docs/Responses/collectionWithItems.md
```diff
@@ -0,0 +1,104 @@
...
+| --- | --- | --- | --- | --- |
+| `id` | `string` | **Yes** |     | Id of the collection |
+| `name` | `string` | **Yes** |     | Name of the collection |
+| `comment` | `string` | No|     | A commnet |
+| `owner` | `string` | No |     | Owner of the collection |
+| `public` | `boolean` | No|     | Is the collection public |
+| `itemCount` | `int` | **Yes** |     | number of items |
+| `created` | `string` | **Yes** |     | Creation date [ISO 8601] |
+| `changed` | `string` | **Yes** |     | Last changed date [ISO 8601] |
+| `allowedUser` | Array of `string` | No |     | A list of allowed usernames |
```
kgarner7: (Same question here about `allowedUser`)

content/en/docs/Endpoints/createcollection.md
```diff
@@ -0,0 +1,144 @@
...
+Creates a collection.
+This endpoint must be accessed using an HTTP POST request.
+
+### Request Body
+
+The request payload should be provided in the body as a JSON object.
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `name` | Yes |  |  | The human-readable name of the collection. |
```
kgarner7: Yes should be bold

content/en/docs/Endpoints/createcollection.md
```diff
@@ -0,0 +1,144 @@
...
+      "id": "200000021"
+    },
+    {
+      "type": "genre",
+      "name": "vaporwave"
+    }
+  ]
+}
+{{< /tab >}}
+{{< /tabpane >}}
```
kgarner7, line 42: Payload should be documented as a separate page. See transcoding extension as an example.

content/en/docs/Endpoints/getcollection.md
```diff
@@ -0,0 +1,117 @@
...
+`http://your-server/rest/getCollection`
+
+Returns a collection with the items it contains.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `id` | **Yes** |  |  | ID of the collection to return, as obtained by `getCollections`. |
+| `itemCount` | No |  |  | The number of items from the collection to return. Max 500. |
```
kgarner7: Why limit the total size? `getPlaylist` for example allows returning the _entire_ payload (including for extremely large playlists).

content/en/docs/Endpoints/getcollection.md
```diff
@@ -0,0 +1,117 @@
...
+| `itemOffset` | No |  |  | The number of items to skip. |
```
kgarner7, line 22: What are the default values for these two fields? 500 and 0?

Is the expected behavior if fetching `itemOffset` outside the range of the collection length (e.g., if the collection is modified while querying) to return nothing?

content/en/docs/Endpoints/getcollections.md
```diff
@@ -0,0 +1,71 @@
...
+
+`http://your-server/rest/getCollections`
+
+Returns all collections a user has access to.
+
+### Parameters
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `username` | No |  |  | If specified, return collections for this user rather than for the authenticated user. The authenticated user must have admin role if this parameter is used. |
```
kgarner7, line 20: Any plans to add pagination, filters, and/or search? If a user makes a large number of collections, it could be messy with no pagination options.
If pagination is offered, should probably include a total count in the response.

content/en/docs/Endpoints/updatecollection.md
```diff
@@ -0,0 +1,92 @@
...
+### Request Body
+
+The request payload should be provided in the body as a JSON object.
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `collectionId` | **Yes** |  |  | The collection ID. |
+| `name` | No |  |  | The human-readable name of the collection. |
+| `comment` | No |  |  | The collection comment. |
+| `public` | No |  |  | `true` if the collection should be visible to all users, `false` otherwise. |
```
kgarner7: Does `public=undefined` (not specified) _also_ make the collection private? It would only make sense if that was `public=false` (with public not being specified having no effect)

content/en/docs/Endpoints/updatecollection.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Request Body
+
+The request payload should be provided in the body as a JSON object.
+
+| Parameter | Req. | OpenS. | Default | Comment |
+| --- | --- | --- | --- | --- |
+| `collectionId` | **Yes** |  |  | The collection ID. |
+| `name` | No |  |  | The human-readable name of the collection. |
+| `comment` | No |  |  | The collection comment. |
```
kgarner7: What happens if these fields are not specified? No change, or clear these fields?
If empty = no change, is there a way to clear the name/comment?

content/en/docs/Endpoints/updatecollection.md
```diff
@@ -0,0 +1,92 @@
...
+| `public` | No |  |  | `true` if the collection should be visible to all users, `false` otherwise. |
+| `add` | No |  |  | Add the specified items to the collection. The payload is an array of [collectionItemID](../payloads/collectionitemid) objects. Items are appended to the tail of the collection. |
+| `move` | No |  |  | Move items from the original position to a new one in the collection. The original range is specified by the `fromStart` (inclusive) and `fromEnd` (exclusive) positions. The first item in the range moves to the position specified by `to` |
+| `remove` | No |  |  | Remove the items at the specified positions from the collection. The payload is an array of integer indices. |
+
+Only one of `add`, `move`, and `remove` can be non-empty in a single request.
```
kgarner7: What's the error for inconsistent payloads?

content/en/docs/Payloads/CollectionItemID.md
```diff
@@ -0,0 +1,42 @@
...
+{
+  "type": "genre",
+  "name": "vaporwave"
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `type` | `string` | **Yes** |     | One of: song, album, artist, playlist |
```
kgarner7: This definition does not include all possible types.

content/en/docs/Payloads/CollectionItemID.md
```diff
@@ -0,0 +1,42 @@
...
+| `id` | `string` | No |     | ID of the item, where applicable |
+| `name` | `string` | No |     | Name of the item, where applicable |
```
kgarner7: What determines whether `id` or `name` is appropriate? Would `id` globally be better (and interpreting the genre value as the `id`)?

content/en/docs/Responses/collection.md
```diff
@@ -0,0 +1,48 @@
...
+| Field |  Type | Req. | OpenS. | Details |
+| --- | --- | --- | --- | --- |
+| `id` | `string` | **Yes** |     | Id of the collection |
+| `name` | `string` | **Yes** |     | Name of the collection |
+| `comment` | `string` | No|     | A commnet |
+| `owner` | `string` | No |     | Owner of the collection |
+| `public` | `boolean` | No|     | Is the collection public |
+| `itemCount` | `int` | **Yes** |     | number of items |
+| `created` | `string` | **Yes** |     | Creation date [ISO 8601] |
+| `changed` | `string` | **Yes** |     | Last changed date [ISO 8601] |
```
kgarner7, line 39: I'm going to assume that `changed` is only updated then `updateCollection` (or `createCollection`) is called, _or_ for server-defined collections (that seems to be implied as an option given the `readonly` and `validUntil` fields) is updated? Or should a server `validUntil` collection not updated the `changed` field?

content/en/docs/Responses/collectionItem.md
```diff
@@ -0,0 +1,281 @@
...
+{{< tab header="Genre" lang="json">}}
+{
+  "genre": {
+    "value": "vaporwave",
+    "songCount": 6,
+    "albumCount": 1
+  }
+}
+{{< /tab >}}
+{{< /tabpane >}}
```
kgarner7, line 331: Insufficient documentation of `collectionItem` (in table format).

kgarner7 (changes requested):
Only reviewed the markdown so far.

content/en/docs/Responses/collection.md
ygabuev: This got carried over from Playlists, for which I also don't see configurations options. I would drop it for now unless there is an argument to keep it in.

content/en/docs/Endpoints/updatecollection.md
ygabuev: Only the specified fields will have an effect on the collection. There should be a distinction between the field being set to the type's default value and it being unset, so that clearing the comment would be done via `{"comment": ""}`.

For example, in Go this would imply unmarshalling into a `*string` type.

I don't think it's valid to clear the name.

content/en/docs/Endpoints/getcollection.md
ygabuev: I agree with your comment above, if `itemCount` is unset, let's return all items. For `itemOffset`, it is 0.

If `itemOffset` is out of range, let's return an empty list?

content/en/docs/Endpoints/createcollection.md
ygabuev: Not sure about that. This follows the `createPlaylist` example - we can deviate from it, but I don't see a big point. An alternative could be to return the `id` of the new collection.

content/en/docs/Endpoints/getcollections.md
ygabuev: Added the pagination, but not sure about filtering and search at this point. Looks like the `search3` endpoint would be a natural place to use, WDYT?

Tolriq: We don't talk about `search3` :)

content/en/docs/Endpoints/updatecollection.md
ygabuev: Since only the fields that are provided have an effect, `public=undefined` keeps the existing value.

content/en/docs/Endpoints/updatecollection.md
ygabuev: Should be 400. Added some basic docs.

content/en/docs/Payloads/CollectionItemID.md
ygabuev: There is a small description above. If `id` exists, it should be `id`. If not, then `name`. Not sure if using `id` for a genre would be clear, but I agree the proposed approach is a bit more complex.

content/en/docs/Endpoints/getcollections.md
ygabuev: Welp :shrug: 

Then what is the current state of search in the community?

Tolriq: It's to be defined, but it's per endpoint with filtering and ordering as kgarner said we also need to define field selections.

content/en/docs/Responses/collection.md
ygabuev: `validUntil` is another carry-over from Playlists, I am not sure how the actual clients and servers use it, do you know? Otherwise I would remove it.

I didn't think about server-defined collections, the `readonly` flag makes sense in the context of one user sharing a collection with another.

ygabuev:
Another thing that's missing atm is something like `getCollectionsByItemID`, which might be helpful during the add-to-collection dialogs.

Also not clear if duplicate items should be allowed.

content/en/docs/Endpoints/createcollection.md
kgarner7: The big point is, if you are creating large collections but _don't_ care about the results, you are then forcing a large serialization step not requested by the client.

content/en/docs/Responses/collection.md
kgarner7: I _suspect_ `allowedUser` is something that was configurable via OG Subsonic **native** api and not documented. If you make it an option on create, that might be good? Dropping it would also be fine.

content/en/docs/Endpoints/getcollection.md
kgarner7: Sounds good. For out of range (at least, in the positive direction), that _could_ happen if you have two interleaving requests (one fetching the collection, paginated, and another removing items from the collection). Of course, the client might _still_ be confused in that case, but maybe an empty would be better.

content/en/docs/Payloads/CollectionItemID.md
kgarner7: I think having consistent `id` is simpler (especially for generators); for `genre`, the name is its identifier after all.

content/en/docs/Responses/collection.md
kgarner7: `validUntil` for playlists is used for server-generated temporary playlists. As an example, suppose you have a "daily jams" server-generated. That should only be valid for the day, so validUntil could say "this is valid until the end of the day".
If you envision served-defined playlists, then keeping this field might make sense. If not, feel free to omit!

content/en/docs/Endpoints/createcollection.md
ygabuev: So wdyt about returning an id or an object with an id?

kgarner7:
> Another thing that's missing atm is something like `getCollectionsByItemID`, which might be helpful during the add-to-collection dialogs.
> 
> Also not clear if duplicate items should be allowed.

For your consideration, playlists allow duplicate items. If you _don't_ allow duplicates, then you should consider whether duplicate IDs should be an _error_, or silently ignored.

content/en/docs/Endpoints/createcollection.md
kgarner7: Thanks for the point! The simplest would be something like `includeItems=false`, which has every field _except_ items (makes the types easy).

content/en/docs/Responses/collection.md
ygabuev: Understood, I don't see it for collections for now, so I'll remove it.

content/en/docs/Endpoints/createcollection.md
kgarner7: An aside for this point, but you _could_ also add that for `getCollection` as well (for example, a client _only_ wishes to get collection _metadata_ and not the items).

lachlan-00:
as a list of items a collection should allow duplicates unless explicitly set

probably as a true/false property of a collection (allowDuplicates / unique)

content/en/docs/Endpoints/createcollection.md
ygabuev: If we do that, we can also combine the `collection` and `collectionWithItems` types with the `items` field optional.

AlexGustafsson:
It's a bit difficult to track the full conversation, but I think I agree with @aweathe.

The main feature I see for this is to be able to group albums, playlists and artists (think discoverability through "daily mixtapes", "artists you like" and things like that).

Mixing playlists and songs is just a bigger playlist. Mixing albums and songs is mostly a playlist with all the albums' songs and the other songs? A collection of songs really is just a playlist. Other combination sounds more of a concern for an application, not the API? What would a portion of an app look like using a collection of songs and cover art, is some sort of precedence?

An "everything API" always has the risk to make client support more difficult and it can lead to differences in implementations (as in, some apps might end up having to ignore one or more of the types of the collection).

As always, it can be a bit hard to convey the intended tone in text, so let me be clear that this mostly sounds like a great feature that I very much look forward to, but I have some concerns and I'm trying to understand the different use cases.

Tolriq:
I don't need this so other don't ? Either we enable mixed content for all and it's generic collections usable for many purpose or we are back to playlist of albums and playlist of artists.

There's no need to add a concept of collection to extend playlists ... Just extend playlists.

jatin-gurjar-enjoy:
i have nothing to add other than the fact that i check this discussion every week because i really do want this feature

Tolriq:
I'd love that too :) But I'm only client side we need reactions from the servers for such features.

jatin-gurjar-enjoy:
> I don't need this so other don't ? Either we enable mixed content for all and it's generic collections usable for many purpose or we are back to playlist of albums and playlist of artists.
> 
> There's no need to add a concept of collection to extend playlists ... Just extend playlists.

I have one question about this which is that but playlist list every song of the albums, while collections don't, allowing for easier management of albums,artists etc by grouping them together.

not that your concept is wrong, yours already allows what i speak of here, but was just wondering really.

Tolriq:
As said a couple of times if we want playlists of albums and artists we expand the already working concept of playlists to support albums and artists to have that.

Collections are another concept of MIXED content, hence the proposal. The added benefits of collections is that they can also work as playlist if you want to use them as that. Filling all the need instead of building on the currently limited playlist api.

content/en/docs/Payloads/CollectionItemID.md
ygabuev: Simplified now :+1:

content/en/docs/Endpoints/createcollection.md
ygabuev: I've modified the `createCollection` endpoint to return the `Collection` type, not `CollectionWithItems`. The latter is now only returned via `getCollection`.

lachlan-00:
2 questions i have left on this.

Duplicates are allowed? we can or can't but i have done this in the Ampache API with unique keys in the database so i assumed no dupes

Is the order or items the index of the list? so all items are displayed in 'track' order.

I've done this in draft in Ampache so i will change a few things based on those 2 things as it's still WIP

ygabuev:
@AlexGustafsson @aweathe @epoupon just my 5c on the concept at hand, I came to implementing this proposal because I want to have collections of albums as a distinct UX concept. For my personal use cases, it specifically don't think of those as larger playlists, and now in my personal collection I don't use playlists as a replacement. I need the album covers, both the grid and the list view, the possibility to order, to export a cohesive JSON, to easily rearrange things. I don't need to see the tracks, and I don't want to accidentally rearrange the tracks or delete them from the list. I also don't see how or why would an application reuse the playlist concept to cover those needs.

The multimedia collection concept is an extension that works for me and that covers more possible use cases, e.g. genre or topic introductions, "best of" lists, selections of work from a particular artist (e.g. solo albums, track features, maybe videos?). Mixed content would bring some UX challenges, but I can't think of them as too painful.

---

@lachlan-00 I am not sure about the duplicates, I don't personally see a need for them and am weakly learning towards disallowing them, unless someone has a good use case. If they are allowed, this might raise UX problems when adding items to collections. If a user decides this on a per-collection basis, there are more stories to cover, both on the server and client side. Not sure.

As for the indices, an item's position in the list is its index. This is already used in the API call `updateCollection` when items are moved or deleted.

lachlan-00:
> @AlexGustafsson @aweathe @epoupon just my 5c on the concept at hand, I came to implementing this proposal because I want to have collections of albums as a distinct UX concept. For my personal use cases, it specifically don't think of those as larger playlists, and now in my personal collection I don't use playlists as a replacement. I need the album covers, both the grid and the list view, the possibility to order, to export a cohesive JSON, to easily rearrange things. I don't need to see the tracks, and I don't want to accidentally rearrange the tracks or delete them from the list. I also don't see how or why would an application reuse the playlist concept to cover those needs.
> 
> The multimedia collection concept is an extension that works for me and that covers more possible use cases, e.g. genre or topic introductions, "best of" lists, selections of work from a particular artist (e.g. solo albums, track features, maybe videos?). Mixed content would bring some UX challenges, but I can't think of them as too painful.
> 
> ---
> 
> @lachlan-00 I am not sure about the duplicates, I don't personally see a need for them and am weakly learning towards disallowing them, unless someone has a good use case. If they are allowed, this might raise UX problems when adding items to collections. If a user decides this on a per-collection basis, there are more stories to cover, both on the server and client side. Not sure.
> 
> As for the indices, an item's position in the list is its index. This is already used in the API call `updateCollection` when items are moved or deleted.

Thanks that makes sense to me.

I do prefer uniqueness on lists as well but I have made it optional for now in my implementation. I would like it to be unique though or at least clear about that in the wording.

I had a look at implementation for Ampache's opensubsonic vs what I've done in the Ampache API and I'm fine with this implementation if we decide on uniqueness too.

content/en/docs/Endpoints/getcollections.md
ygabuev: Added a simple filter by name and sorting.

jatin-gurjar-enjoy:
:)

ygabuev:
@lachlan-00 added some explications that mandate item uniqueness. Happy to change that later if someone proposes a strong argument against.

PR #242, opened 2026-05-22
feat: add tinysub to client list

intergrav:
https://devins.page/tinysub

### DIFF
--- a/content/en/docs/_index.md
+++ b/content/en/docs/_index.md
@@ -50,7 +50,7 @@ Any server or client can join the organization and make proposals for [OpenSubso
 | [Supersonic](https://github.com/dweymouth/supersonic)  |   |
 | [Symfonium](https://symfonium.app/)  | [Documentation](https://support.symfonium.app/t/1178) |
 | [Tempus](https://github.com/eddyizm/tempus)  |  |
-| [tinysub](https://devins.page/tinysub)  |  |
+| [tinysub](https://devins.page/tinysub)  | [Documentation](https://tangled.org/devins.page/tinysub#opensubsonic-extensions) |
 
 ## Goals
 
--- a/content/en/docs/_index.md
+++ b/content/en/docs/_index.md
@@ -50,6 +50,7 @@ Any server or client can join the organization and make proposals for [OpenSubso
 | [Supersonic](https://github.com/dweymouth/supersonic)  |   |
 | [Symfonium](https://symfonium.app/)  | [Documentation](https://support.symfonium.app/t/1178) |
 | [Tempus](https://github.com/eddyizm/tempus)  |  |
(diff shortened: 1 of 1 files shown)


Tolriq (dismissed):

kgarner7 (dismissed):

paulijar:
Being curious,  and as there's no documentation link provided, which of the OpenSubsonic enhancements does tinysub support?

Note that some of the OpenSubsonic servers don't support the token+salt authentication. I don't think it's any strict requirement for OpenSubsonic clients but it would sure be nice if the client could support either user+password authentication or the apiKey authentication.

intergrav:
i can add support for user+pass and api key auth :) i'll turn this PR into a draft for now while i work on that

tinysub supports these additions and extensions currently:

- ReplayGain
- Song Lyrics

i plan on also supporting these extensions in the future:

- Index based Queue
- Sonic similarity
- Transcode offset

intergrav:
i've added support for cleartext auth as a fallback if token auth is unsupported on the server

i've also added support for API key auth, however i have no way to test it myself *(as i assume neither Navidrome or Gonic support it, which are the only servers i'm running atm)*. i just implemented it according to specification, so it should work, but once again no way to test myself

i've also added support for the index based queue extension and formpost detection, and i'll possibly look into sonic similarity and transcode offset extensions later if i get the time to *(aka feel motivated enough)*. i put a little section at the bottom of tinysub's readme about the supported extensions, so i'll add a link to that in the documentation column in a sec

Tolriq (approved):

paulijar (approved):

paulijar:
Thanks. I tested both user-password and the apiKey authentication with Nextcloud Music. I could successfully log in with both mechanisms, but for some reason, the playlists didn't show up when the apiKey was being used. This is weird since I can see that the app makes a valid request for `getPlaylists` and gets the usual response.

intergrav:
> I could successfully log in with both mechanisms, but for some reason, the playlists didn't show up when the apiKey was being used. This is weird since I can see that the app makes a valid request for `getPlaylists` and gets the usual response.

pushed an update that should hopefully fix this, could you try again? (may need to do a hard reload to update)

if it's still broken, i'll just mark it as a WIP feature for now, and i'll see if i can spin up a server with API key support to experiment with at some point

paulijar:
> pushed an update that should hopefully fix this, could you try again? (may need to do a hard reload to update)

Tested again and worked great after the hard reload. 🎉

intergrav:
subsonic transcoding and opensubsonic transcodeOffset are now supported, added to readme

lachlan-00 (approved):

lachlan-00:
@intergrav love the project, looks super cool!

PR #257, opened 2026-07-16
Add `ArtistID3.disambiguation` field

sentriz:
Add an optional `disambiguation` field to the ArtistID3 response element, a free-text comment used to distinguish between artists sharing the same name (e.g. MusicBrainz-style "French electronic duo" vs "drum & bass producer")

based on https://github.com/opensubsonic/open-subsonic-api/discussions/249

### DIFF
--- a/content/en/docs/Responses/ArtistID3.md
+++ b/content/en/docs/Responses/ArtistID3.md
@@ -20,6 +20,7 @@ description: >
   "starred": "2017-04-11T10:42:50.842Z",
   "musicBrainzId": "189002e7-3285-4e2e-92a3-7f6c30d407a2",
   "sortName": "Mello (2)",
+  "disambiguation": "French electronic duo",
   "roles": [
     "artist",
     "albumartist",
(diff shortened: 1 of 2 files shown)


paulijar (dismissed):

kgarner7 (dismissed):

Tolriq:
Isn't the MBID exactly for that ? How are servers supposed to fill that and detect that and clients use it and display it ?

kgarner7:
Disambiguation is a _visual_ alias that can be displayed for the user. For example, you could have two artists named `A`, but one is a vocalist, and another is a composer. The client could then show `A (vocalist)` and `A (composer)`, for example.

Tolriq:
Ok then we must make it clear that it's not only internal disambiguation but something clients should handle / display so that servers don't put random stuff there.

sentriz:
> Isn't the MBID exactly for that ? How are servers supposed to fill that and detect that and clients use it and display it ?

For clients displaying it, it's easy really. Many of client's "Artists" pages are fed straight from getArtists, so they just read the field and show it when listing the artists

For servers, at least for gonic, it will compare with info it has fetched from MusicBrainz when the artist has a MusicBrainz ID. the same system/table that feeds info for getArtistInfo for example

sentriz:
> Ok then we must make it clear that it's not only internal disambiguation but something clients should handle / display so that servers don't put random stuff there.

Sure makes sense, updated. wdyt?

sentriz:
thanks for the review @paulijar @kgarner7 would you mind having another look? got dismissed after last push

Tolriq (approved):

Tolriq:
Looks clear enough.

epoupon (approved):

kgarner7 (approved):

sentriz:
thanks!

PR #258, opened 2026-07-16
Add extended library browsing extension

Tolriq:
WIP to show the concept and why it's the proper solution to address all needs instead of adding small temporary double work only solving a small part of the problem.


sentriz:
@Tolriq would Synfonium use only these endpoints for syncing instead of search3, for servers that support it?

Tolriq:
Yes that's one of the goal, but the main gain is for online clients that have proper access to all the data efficiently.

The main gain for Symfonium (and offline clients) will be in the second phase with the addition of an official lastModified for all data with sort and filter on it (on user data change for example, or metadata) + a database status to be aware of last changes, deletion and non possible incremental sync (As it is was added in Kodi a long time ago).

That would allow differential sync and be a huge gain.

jeffvli:
Not a fan of the cursor-based pagination approach as it does not solve the problems I had with the original list endpoints.

The original list endpoints allowed you to paginate via limit/offset, but did not return a total. Despite being able to jump to specific pages via offset, the client could not determine how many pages there actually were without first fetching the entire list.

The proposal now includes a total in the response, but using cursors means that the client cannot utilize it as the API restricts users to only jumping from one page to the next.

Tolriq:
Cursors are needed for sorts like random. If you have better solution that covers all cases and servers are willing to implement propose, this is an open discussion until we try to actually do something when I'm back.

jeffvli:
The solution I saw proposed by @deluan for random sorting was using offset based pagination with a deterministic seed that can either be provided from or passed to the server.

Tolriq:
This does not really cover the other case of cursors being speed. We can have the api return an optional queryId and if present the client must send it again for the other pages. Covers both needs.

jeffvli:
Would it be too much to ask that servers implement both offset AND cursor based pagination to let clients pick and choose as needed? I understand the want for cursors as it would allow for faster offline-sync, but you *did* mention that the main goal is for enhancements for online clients.

Having experienced some of my most-used online services switching from offset to cursor based pagination, I have found that (at least for desktop usage) cursor pagination has ultimately decreased my usability for those services. It makes sense to use it at scale due to speed and cost-savings, but I would say the majority of OpenSubsonic users are running their servers for personal usage. In that case, UX is more important than a gain in performance.

Tolriq:
Speed matters for online too ;) Cursors also allow faster next page for online on slow queries. You can have complex sort and filters here, speed matter for normal usage.

We can add start index that could force drop the cursor and create a new one. But not sure who needs to skip to page 43 without knowing what's there, usually they refine the filter.

jeffvli:
I'm going to stand by my suggestion that the spec should implement both cursor and offset pagination. Unfortunately without offsets, I would not be willing to implement this for my client.

Tolriq:
> We can add start index that could force drop the cursor and create a new one.

jeffvli:
I would be fine with the start index solution.

Tolriq:
@opensubsonic/servers we need your participation on this to know if you are onboard or not for such change as it can be a little work on your side for the filter part.

content/en/docs/Extensions/extendedLibrary.md
```diff
@@ -0,0 +1,462 @@
...
+
+Logical arrays must contain at least one node. A node mixing `field`, `and`, or
+`or` shapes is invalid.
+
+### Operators and values
+
+| Family | Operators | Value |
+| --- | --- | --- |
+| Equality/set | `equals`, `notEquals`, `in`, `notIn` | One typed value, or a non-empty homogeneous array for set operators. |
+| Text/list | `contains`, `notContains`, `startsWith`, `endsWith` | A string. |
```
gravelld, line 156: "matches" for regexes? Or maybe "contains" is a contains-regex...?

Tolriq: No plan for regex that would be a real burden for servers depending on what regex and what database and performances impact.

content/en/docs/Extensions/extendedLibrary.md
```diff
@@ -0,0 +1,462 @@
...
+`sort` is an ordered array of one to five objects with `field` and `direction`.
+Directions are `ascending` and `descending`, and each field may appear only
+once.
+
+If `sort` is omitted, the effective sort is `id ascending`. If `id` is
+explicit, it must be the last key. Otherwise the server appends `id ascending`
+as a deterministic tie-breaker. Missing values always sort after present
+values, for both directions.
+
+### Sortable song fields
```
gravelld, line 244: Is there a reason some fields are left out - just phasing?

content/en/docs/Extensions/extendedLibrary.md
```diff
@@ -0,0 +1,462 @@
...
+  "cursor": "opaque-next-cursor"
+}
+```
+
+Servers must bind and integrity-protect a cursor against the authenticated
+principal, endpoint, filter, and sort. Invalid, expired, cross-user, and
+mismatched cursors fail with generic error code `0`. Cursors are not portable
+between the three entity browsing endpoints or the facet endpoint.
+
+Cursors do not create a snapshot. `total` is recomputed from the currently
```
gravelld, line 348: Do we want to be more vague on this - maybe a server implementation may wish to create a snapshot. For example, the "server processes" list below is possibly expensive and some server implementers may not want to do it this way.

content/en/docs/Extensions/extendedLibrary.md
Tolriq: This was a quick WIP to expose the concept.

content/en/docs/Extensions/extendedLibrary.md
Tolriq: ?

content/en/docs/Extensions/extendedLibrary.md
```diff
@@ -0,0 +1,462 @@
...
+    {"field": "album", "direction": "ascending"}
+  ],
+  "limit": 50,
+  "cursor": "opaque-next-cursor"
+}
+```
+
+Servers must bind and integrity-protect a cursor against the authenticated
+principal, endpoint, filter, and sort. Invalid, expired, cross-user, and
+mismatched cursors fail with generic error code `0`. Cursors are not portable
```
gravelld, line 345: Is it not time to add more specific error codes? A generic "Bad request" style error?

gravelld:
Are these changes discussed somewhere to understand some of the backstory/thinking? I've been reading the [extension docs](https://github.com/Tolriq/open-subsonic-api/blob/03c6a25a38376785c6e36ec12fea76e77b481441/content/en/docs/Extensions/extendedLibrary.md). I'm taking that as the entry point.

The consistency-forced naming and duplication with legacy operations is obviously undesirable but understandable to make progress. I do wonder if one day we will need a proper 2.0 with these things sorted, because new consumers might be a bit baffled about which operations to use early on.

Overall I think this appears a really well thought-through spec.

content/en/docs/Extensions/extendedLibrary.md
Tolriq: Yes we can.

content/en/docs/Extensions/extendedLibrary.md
gravelld: Let me know if there was something that was unclear.

Tolriq:
@deluan @epoupon

epoupon:
> @deluan @epoupon

Hello!

I'm completely in favor of these changes!
There's quite a bit of work involved, so I'm going to do it iteratively: I'll add the pieces little by little in LMS to bring it closer to this proposal (being able to filter on more things, being able to combine filters, operators, etc.)

deluan:
I already have almost all of these features implemented in Navidrome, so no concerns. But I kind of agree with @gravelld that this has too much clash with OG Subsonic API, and can cause much confusion for new devs. Maybe it is time to discuss v2.0?

Tolriq:
How do you see a v2 that could be supported by more than just LMS and you ? Redefining everything would limit a lot servers implementation versus slowly adding new features and phasing some out as clients start to stop supporting them. We can try to make a full table with servers and clients as we did to create OS but seeing current status I doubt we can reach something in under 10 years.

PR #259, opened 2026-07-17
Add `topSongsByArtistId` extension

sentriz:
`getTopSongs` currently only accepts an artist **name**. Which is ambiguous since two
artists can share the same name, so the server has to guess which one the client
meant

Passing an artist **ID** removes the ambiguity 

- A server backed by something like ListenBrainz (MBID based) can resolve the
  exact artist directly, disambiguating same named artists
- Even a string-based backend like last.fm benefits. The server can resolve the
  artist from the ID it already knows, then fetch/filter that artist's top songs
  precisely instead of round-tripping a name

### DIFF
--- a/content/en/docs/Endpoints/getopensubsonicextensions.md
+++ b/content/en/docs/Endpoints/getopensubsonicextensions.md
@@ -56,6 +56,12 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a nested
                 "versions": [
                     1
                 ]
+            },
+            {
+                "name": "topSongsByArtistId",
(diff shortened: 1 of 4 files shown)


Tolriq:
For the record this also fit https://github.com/opensubsonic/open-subsonic-api/pull/258 ;) filter artistId sort by topSongs and is one more argument for it ;)

paulijar (approved):

kgarner7 (approved):

paulijar (approved):

deluan (approved):

PR #261, opened 2026-07-19
feat: add public image url extension

kgarner7:
For clients, `getCoverArt` is completely unsuitable for sharing publicly (e.g., for discord rpc). This extension provides a way to convert a request for `/rest/getCoverArt` to a URL that could be shared publicly.

### DIFF
--- a/content/en/docs/Endpoints/getpublicimageurl.md
+++ b/content/en/docs/Endpoints/getpublicimageurl.md
@@ -1,0 +1,56 @@
+---
+title: "getPublicImageURL"
+linkTitle: "getPublicImageURL [OS]"
+opensubsonic:
+- Addition
+categories:
+- Media retrieval
+description: >
+  Returns a URL for accessing cover art publicly.
+---
+
+`http://your-server/rest/getPublicImageURL`
+
+Returns a cover art image.
+
+### Parameters
+
+| Parameter | Req.    | OpenS. | Default | Comment |
+| --------- | ------- | ------ | ------- | --- |
+| `id`      | **Yes** |        | The coverArt ID. Returned by most entities likes [`Child`](../../responses/child) or [`AlbumID3`](../../responses/albumid3) | |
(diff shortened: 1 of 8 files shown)


Tolriq:
How would the discord RPC use the url ? Many people are behind proxy with headers auth, vpn, cloudflare tunnels, and others cases that would not work. Usually the client proxy or download, this is already needed to cast to Chromecast and have covers as UPnP.

I don't think the server can know if those urls are actually accessible or not. Navidrome IIRC have tons of settings to workaround some cases but still.

dweymouth:
> How would the discord RPC use the url ? Many people are behind proxy with headers auth, vpn, cloudflare tunnels, and others cases that would not work. Usually the client proxy or download, this is already needed to cast to Chromecast and have covers as UPnP.
> 
> I don't think the server can know if those urls are actually accessible or not. Navidrome IIRC have tons of settings to workaround some cases but still.

It wouldn't work in all cases, for example cloudflare tunnels as you mentioned, but I see the value for the case of users with publicly-accessible servers e.g. behind a simple reverse proxy with HTTPS. With Chromecast and uPnP the client can serve covers and media to the renderer since they're on the same LAN, but a client can't open a public address that Discord's servers can reach out to for the image. Only a publicly accessible server can do that.

lachlan-00 (approved):

epoupon:
So, if I understand correctly, this would require servers to support a new unauthenticated endpoint that can:
- generate unique temporary links
- include a configuration option to specify the actual deployment path

Servers should allow this on a per-user basis, probably disabled by default.

Tolriq:
And currently no control on the client about the expiration and to what format it's resized if the source is an animated image for example.

The thing is that I think having a proper "transcoding/resizing" endpoint for images would be nice for clients too, but having to make 1 call per image makes this slow.

In the end this is an extension point just because of discord that does not support uploading the image and requires a public url, can't we make this more useful ?

PR #262, opened 2026-07-25
feat(songLyrics v2): add explicit line end timing

ranokay:
## Summary

`line.start` tells a client when a lyric line begins, but it cannot describe when the final line ends or distinguish a deliberate gap, an overlap, and an instantaneous marker. This adds an optional `line.end` value so enhanced lyrics clients can represent those timings without inferring them from the next line.

## Compatibility

- `end` uses the same track-relative millisecond timeline as `start`; the structured-lyrics offset applies to both values.
- It requires `start`, must be greater than or equal to it, and may equal it for an instantaneous marker. Gaps and overlaps remain valid.
- An omitted `end` means that the end is unknown. It does not imply the next line's start.
- The field is available only in `enhanced=true` responses. Unsynced lyrics omit both timing values, while default responses and the advertised `[1, 2]` versions remain unchanged.

### DIFF
--- a/content/en/docs/Endpoints/getLyricsBySongId.md
+++ b/content/en/docs/Endpoints/getLyricsBySongId.md
@@ -123,7 +123,7 @@ Does not exist.
 
 #### Version 2 (`enhanced=true`)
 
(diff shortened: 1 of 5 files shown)


Tolriq:
```Its omission means the end is unknown; clients **must not** assume the line lasts until the next line starts.``` 

If there's no end clients needs to assume things, so the wording is probably too strong.

ranokay:
You're right. I was trying to distinguish an exact, source-provided end from an end inferred by the client, but yeah, `must not` is too restrictive.
I'll change the wording to:
`Its omission means the exact end is unknown. Clients may infer a fallback end, for example from the next line's start, but should not treat it as exact source timing.`

Thanks for pointing that out.

Tolriq (approved):

PR #265, opened 2026-08-07
missing data TODOs

lachlan-00:


lachlan-00:
@opensubsonic/servers  and @opensubsonic/clients  can you review and get this one in? the spec was missing a fair bit of info when i was validating the spec

content/en/docs/Endpoints/getvideoinfo.md
```diff
@@ -28,13 +28,45 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a nested
...
+{
+  "subsonic-response": {
+    "status": "ok",
+    "version": "1.16.1",
+    "type": "ampache",
+    "serverVersion": "8.0.0",
+    "openSubsonic": true,
+    "videoInfo": {
+      "id": "vi-1"
+    }
```
paulijar, line 53: Could there be some more fields than just the `id` also in this JSON example, like there's in the XML example? Even in case there's nothing to show here, having the empty arrays for `captions`, `audiotTrack`, and `conversion` in the JSON format would make sense. However, the OG Subsonic seems to omit these empty arrays as can be seen in http://demo.subsonic.org/rest/getVideoInfo.view?u=guest&p=guest&v=1.12.0&c=myapp&f=json&id=460.

content/en/docs/Responses/videos.md
```diff
@@ -8,12 +8,96 @@ description: >
...
+      "path": "SPACE_Magic_Fly_1977.mp4",
+      "originalWidth": 640,
+      "originalHeight": 480
+    }
+  ]
+}
 {{< /tab >}}
 {{< tab header="Subsonic" lang="json" >}}
-// TODO
+"videos" : {
```
paulijar: I believe the key "videos" is not supposed to be shown here. It's not there in the OpenSubsonic version and the convention in other similar pages too seems to be to show the response type as an anonymous JSON object.

content/en/docs/Responses/audioTrack.md
```diff
@@ -0,0 +1,25 @@
...
+
+{{< tabpane persistLang=false >}}
+{{< tab header="**Example**:" disabled=true />}}
+{{< tab header="OpenSubsonic" lang="json">}}
+// TODO
+{{< /tab >}}
+{{< tab header="Subsonic" lang="json" >}}
+{
+  "id" : "83",
+  "bitRate" : 1000
```
paulijar: This is incorrect example for audioTrack

content/en/docs/Responses/audioTrack.md
```diff
@@ -0,0 +1,25 @@
...
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| ----- | ---- | ---- | ------ | ------- |
+| `id` | string | **Yes** | | The ID of a transcoded version of the video file |
+| `name` | string | | | Language Name (e.g. English, French) |
+| `languageCode` | string | | | ISO 639 (2/3) code |
```
paulijar: Please fill the "No" to the "Req." field. (For the "OpenS." column, on the other hand, the convention of the document seems to be to omit the "No" values so that's fine also here).

content/en/docs/Responses/captions.md
```diff
@@ -0,0 +1,24 @@
...
+  "id" : "0",
+  "name" : "Planes 2.srt"
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| ----- | ---- | ---- | ------ | ------- |
+| `id` | string | **Yes** | | The ID of a transcoded version of the video file |
+| `name` | string | | | Subtitle track name |
```
paulijar: Again, the "Req. = No" should be filled to the table.

content/en/docs/Responses/videoConversion.md
```diff
@@ -0,0 +1,25 @@
...
+  "id" : "83",
+  "bitRate" : 1000
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| ----- | ---- | ---- | ------ | ------- |
+| `id` | string | **Yes** | | The ID of a transcoded version of the video file |
+| `bitRate` | int | | | File bitrate |
```
paulijar: It should be mentioned here that the unit of `bitRate` is Kbps (as documented in https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd).

And add the "Req. = No" values.

paulijar (changes requested):
Thanks, this is good an valuable work in filling the gaps.

content/en/docs/Responses/searchResult.md
```diff
@@ -8,12 +8,152 @@ description: >
...
+  } ]
+}
 {{< /tab >}}
 {{< /tabpane >}}
 
-| Field |  Type | Req. | OpenS. | Details |
-| --- | --- | --- | --- | --- |
+| Field       | Type                        | Req.    | OpenS. | Details                       |
+|-------------|-----------------------------|---------|--------|-------------------------------|
+| `match`     | Array of [`Child`](../child) |         |        | Search results                |
```
gravelld, line 157: I don't see any reference that says the `match`es were `Child`ren in the original Subsonic spec, but given the examples given in the old Subsonic examples appear to be subsets of Child records, I guess this is fine.

paulijar: @gravelld This actually is defined in https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd:
image

gravelld: Ah yes, forgot about the xsd... 👍🏽

content/en/docs/Endpoints/getvideoinfo.md
kgarner7: Related, but in the XML version below, you have `captions`, `audioTrack` and `conversion`, but those are not included in the JSON. In general, the XML and JSON should reflect the same object, so copying those over should help here.

content/en/docs/Endpoints/getvideos.md
```diff
@@ -26,13 +26,111 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a nested
...
+            "title" : "SPACE_Magic_Fly_1977",
+            "size" : 20084777,
+            "contentType" : "video/mp4",
+            "suffix" : "mp4",
+            "path" : "SPACE_Magic_Fly_1977.mp4",
+            "isVideo" : true,
+            "playCount" : 1,
+            "created" : "2021-02-23T04:21:29.040Z",
+            "type" : "video"
+         } ]
```
kgarner7, line 139: Is there a reason why the Subsonic version only has two items, while OS has three? Let's keep the data consistent (only adding additional fields as needed).

content/en/docs/Endpoints/search.md
```diff
@@ -36,10 +36,162 @@ A [`subsonic-response`](../../responses/subsonic-response) element with a nested
...
+{
+  "subsonic-response": {
+    "status": "ok",
+    "version": "1.16.1",
+    "type": "ampache",
+    "serverVersion": "8.0.0",
+    "openSubsonic": true,
+    "searchResult": {
+      "offset": 0,
+      "totalHits": 5,
```
kgarner7, line 48: Keep this as 2 to match the actual results?

content/en/docs/Responses/searchResult.md
```diff
@@ -8,12 +8,152 @@ description: >
 {{< tabpane persist=false >}}
 {{< tab header="**Example**:" disabled=true />}}
 {{< tab header="OpenSubsonic" lang="json">}}
-// TODO
+{
+  "offset": 0,
+  "totalHits": 5,
```
kgarner7, line 13: Same here, it might make sense to keep the `totalHits` the same as the results

content/en/docs/Responses/videoInfo.md
```diff
@@ -8,12 +8,20 @@ description: >
...
+  "id": "vi-1"
+}
 {{< /tab >}}
 {{< tab header="Subsonic" lang="json" >}}
-// TODO
+{
+  "id" : "83"
+}
 {{< /tab >}}
 {{< /tabpane >}}
```
kgarner7, line 47: The examples listed should in general have all of the fields populated to be most useful. Feel free to use the same `videoInfo` example(s) across all the docs, but it makes it more useful here.

kgarner7 (changes requested):
A general note: if you have data for XML, Subsonic, and OpenSubsonic, please keep the data _consistent_ across all three domains (including new fields in OS as appropriate). I'd _rather_ have the same XML/JSON object(s) reused over the place (with the appropriate subsets) for consistency.

content/en/docs/Endpoints/search.md
paulijar: Or maybe it was intentional to have a different number here? When the `count` parameter limits the number of results, the `totalHits` may be greater than the number of results shown.

lachlan-00:
I've been in japan over august and sep, back to work today so i'll update

content/en/docs/Responses/videoConversion.md
```diff
@@ -0,0 +1,29 @@
...
+  "bitRate" : 1000
+}
+{{< /tab >}}
+{{< /tabpane >}}
+
+| Field | Type | Req. | OpenS. | Details |
+| ----- | ---- | ---- | ------ | ------- |
+| `id` | string | **Yes** | | The ID of a transcoded version of the video file |
+| `bitRate` | int | No | | File bitrate, in Kbps |
+| `audioTrackId` | int | No | **Yes** | The ID of the audio track this version carries |
```
paulijar: The `audioTrackId` is *not* an OpenSubsonic addition, it can be found from https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd.

paulijar (changes requested):
One more small finding, otherwise looks great now.

paulijar (approved):

PR #270, opened 2026-09-18
Add funoteka to the list of servers

kzntsv-dev:
Adds [funoteka](https://github.com/kzntsv-dev/funoteka) — a self-hosted Subsonic/OpenSubsonic music server (TypeScript, MIT, runs under Docker or natively).

OpenSubsonic support, verified against a running server:

- every response carries the server identification triple: `type="funoteka"`, `serverVersion`, `openSubsonic="true"`;
- `getOpenSubsonicExtensions` answers **without authentication** and reports `indexBasedQueue`, `playbackReport`, `transcodeOffset`, `transcoding` and others.

The row is inserted in the existing alphabetical order (after EdgeSonic, before gonic). The second column is `-` for now, as there is no separate OpenSubsonic documentation page yet — I am happy to add one and link it here if you prefer that.

funoteka's niche is libraries that live in folders: an absolute folder path is the album identity, image files with cue sheets become real playable tracks, and box sets stay a box set plus its disc albums.


paulijar:
Looks good to me but providing the OpenSubsonic documentation link would be preferable. The main point there is to list the OpenSubsonic features implemented, to help the client developers to see what could be tested against each server.

kzntsv-dev:
Thanks — added. The funoteka row now links to [docs/OPENSUBSONIC.md](https://github.com/kzntsv-dev/funoteka/blob/main/docs/OPENSUBSONIC.md).

That page is exactly what you described: the OpenSubsonic methods this server answers, grouped by protocol area, plus the extensions it declares via `getOpenSubsonicExtensions` (with versions), the methods that are empty by design, the ones rejected for byte-format reasons, and the `unknown method` list with the reason for each.

It is generated from the router and a test compares the method names it lists against the actual handlers, so it cannot silently drift out of date.

paulijar (approved):

kgarner7 (approved):

